PR #169, opened 2020-01-21
Users/t anmah/simulation dependency checker fix

andreamah:
# Description:

Fixed automated package installer for Python libraries needed for custom adafruit_circuitplayground.express library. 
 - Fixed pywin32 requirements version for Python 3.8
 - Improved error handling for Python package installation; used to fail silently, but now pops up with an error message. 
 - More robust way of checking whether package installation failed; now checks for presence of ./out/python_libs to see if installation failed.
 - improved README + etc. docs for new dependency workflow (no more need to pip install packages globally).

## Type of change

# Limitations:

Sound functionality still fails on Mac OS. Issue already recorded on backlog item 28661. 

# Testing:

- [ ] Manual testing (test with/without dependencies installed prior, test with bad requirements.txt, tested on Mac OS, and Windows.

# Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,19 +43,15 @@ You will be prompted to install the Python dependencies during the first use.
 - _**[Python 3.7.4](https://www.python.org/downloads/)**_: Make sure you've added python and pip to your PATH in your environment variables. (1)
 - _**[Python VS Code extension](https://marketplace.visualstudio.com/items?itemName=ms-python.python)**_: This will be installed automatically from the marketplace when you install Device Simulator Express.
 
-The following dependecies can be installed for you by the extension by clicking yes when you are prompted to (**except** `pywin32` which is needed only on Windows platform). (2)
+- Python Modules
+  - **Note:** On extension activation, you will be prompted with a popup message asking if you want the modules to be automatically installed for you. The following Python modules should be downloaded when you select "yes" on prompt message. **If modules are not installed correctly, please use the "pip install" commands listed below.**
+    - Playsound : `pip install playsound`
+    - Pywin32 : `pip install pywin32`
+      - **Note:** This is only needed for Windows computers.
+    - Python-Socketio : `pip install python-socketio`
+    - Requests : `pip install requests`
+    - Application Insights: `pip install applicationinsights`
 
-- _**Playsound**_  
-  install by typing the following commands in a console: `pip install playsound`
-
-- _**Pywin 32**_  
-  install by typing the following commands in a console (only for Windows computers, you must run it manually): `pip install pywin32`
-- _**Python-Socketio**_  
-   install by typing the following commands in a console: `pip install python-socketio`
-- _**Requests**_  
-   install by typing the following commands in a console: `pip install requests`
-- _**Application Insights**_  
-  install by typing the following commands in a console: `pip install applicationinsights`
 
 ## Useful Links
 
--- a/README.md
+++ b/README.md
@@ -43,14 +43,11 @@ You will be prompted to install the Python dependencies during the first use.
 - _**[Python 3.7.4](https://www.python.org/downloads/)**_: Make sure you've added python and pip to your PATH in your environment variables. (1)
 - _**[Python VS Code extension](https://marketplace.visualstudio.com/items?itemName=ms-python.python)**_: This will be installed automatically from the marketplace when you install Device Simulator Express.
 
-- Python Modules for Deploying to Device
-    - If you're on Windows, use the following command in the console to install pywin32: `pip install pywin32`
-
 - Python Modules for Simulation
   - **Note:** On extension activation, you will be prompted with a popup message asking if you want the modules to be automatically installed for you. The following Python modules should be downloaded when you select "yes" on the prompt message. **If modules are not installed correctly, please use the "pip install" commands listed below.**
     - Playsound : `pip install playsound`
     - Pywin32 : `pip install pywin32`
-      - **Note:** This is only needed for Windows computers.
+      - On Windows, you need to use the above command in the console to manually install pywin32.
     - Python-Socketio : `pip install python-socketio`
     - Requests : `pip install requests`
     - Application Insights: `pip install applicationinsights`
@@ -100,6 +97,8 @@ Before deploying the python code to your CPX device, you need to format your dev
 1. Download the firmware with the .uf2 file (link: https://learn.adafruit.com/adafruit-circuit-playground-express/circuitpython-quickstart)
 2. Download the lastest version of the cpx library (link: https://learn.adafruit.com/welcome-to-circuitpython/circuitpython-libraries).  
    **_Note:_** Make sure you name your file main.py or code.py: the device automatically runs the first file that is likely named.
+
+Then, if you are on Windows, you will also need to install the Python Pywin32 package. Use the following command in the console: `pip install pywin32`
 
 !["Deploy to Device" example](https://www.microsoft.com/en-us/garage/wp-content/uploads/sites/5/2019/08/deployToBoard.png)
 
--- a/README.md
+++ b/README.md
@@ -43,11 +43,11 @@ You will be prompted to install the Python dependencies during the first use.
 - _**[Python 3.7.4](https://www.python.org/downloads/)**_: Make sure you've added python and pip to your PATH in your environment variables. (1)
 - _**[Python VS Code extension](https://marketplace.visualstudio.com/items?itemName=ms-python.python)**_: This will be installed automatically from the marketplace when you install Device Simulator Express.
 
-- Python Modules for Deploy to Device
+- Python Modules for Deploying to Device
     - If you're on Windows, use the following command in the console to install pywin32: `pip install pywin32`
 
 - Python Modules for Simulation
-  - **Note:** On extension activation, you will be prompted with a popup message asking if you want the modules to be automatically installed for you. The following Python modules should be downloaded when you select "yes" on prompt message. **If modules are not installed correctly, please use the "pip install" commands listed below.**
+  - **Note:** On extension activation, you will be prompted with a popup message asking if you want the modules to be automatically installed for you. The following Python modules should be downloaded when you select "yes" on the prompt message. **If modules are not installed correctly, please use the "pip install" commands listed below.**
     - Playsound : `pip install playsound`
     - Pywin32 : `pip install pywin32`
       - **Note:** This is only needed for Windows computers.
--- a/README.md
+++ b/README.md
@@ -43,7 +43,10 @@ You will be prompted to install the Python dependencies during the first use.
 - _**[Python 3.7.4](https://www.python.org/downloads/)**_: Make sure you've added python and pip to your PATH in your environment variables. (1)
 - _**[Python VS Code extension](https://marketplace.visualstudio.com/items?itemName=ms-python.python)**_: This will be installed automatically from the marketplace when you install Device Simulator Express.
 
-- Python Modules
+- Python Modules for Deploy to Device
+    - If you're on Windows, use the following command in the console to install pywin32: `pip install pywin32`
+
+- Python Modules for Simulation
   - **Note:** On extension activation, you will be prompted with a popup message asking if you want the modules to be automatically installed for you. The following Python modules should be downloaded when you select "yes" on prompt message. **If modules are not installed correctly, please use the "pip install" commands listed below.**
     - Playsound : `pip install playsound`
     - Pywin32 : `pip install pywin32`
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -46,7 +46,7 @@ export const CONSTANTS = {
       "[ERROR] A debugging session is currently in progress, please stop it before running your code. \n"
     ),
     DEPENDENCY_DOWNLOAD_ERROR:
-      "Package downloads failed. Some functionality may not work. Try restarting the simulator or review the installation docs.",
+      "Package downloads failed. Some functionalities may not work. Try restarting the simulator or review the installation docs.",
     FAILED_TO_OPEN_SERIAL_PORT: (port: string): string => {
       return localize(
         "error.failedToOpenSerialPort",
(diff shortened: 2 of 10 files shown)


src/requirements-windows.txt
```diff
@@ -0,0 +1,6 @@
+playsound==1.2.2
+pytest==5.0.1
+applicationinsights==0.11.9
+python-socketio==4.3.1
+requests==2.22.0
+pywin32>=224
```
nasadigital: Do you think it would make sense to fix pywin32's version to 224 in case there are breaking changes in future versions?

README.md
```diff
@@ -43,19 +43,18 @@ You will be prompted to install the Python dependencies during the first use.
...
-
-- _**Pywin 32**_  
-  install by typing the following commands in a console (only for Windows computers, you must run it manually): `pip install pywin32`
-- _**Python-Socketio**_  
-   install by typing the following commands in a console: `pip install python-socketio`
-- _**Requests**_  
-   install by typing the following commands in a console: `pip install requests`
-- _**Application Insights**_  
-  install by typing the following commands in a console: `pip install applicationinsights`
+- Python Modules for Deploy to Device
```
nasadigital: _&ast;nit&ast;_
What do you think of something like "Python Modules for **Deploying** to Device" instead of "Python Modules for **Deploy** to Device"?
If "Deploy to Device" is a command and we want to keep it here, maybe putting it inside of quotes or formatting it as inline code (`Deploy to Device`) might make sense?
Same for the other docs.

README.md
```diff
@@ -43,19 +43,18 @@ You will be prompted to install the Python dependencies during the first use.
...
+    - If you're on Windows, use the following command in the console to install pywin32: `pip install pywin32`
+
+- Python Modules for Simulation
+  - **Note:** On extension activation, you will be prompted with a popup message asking if you want the modules to be automatically installed for you. The following Python modules should be downloaded when you select "yes" on prompt message. **If modules are not installed correctly, please use the "pip install" commands listed below.**
```
nasadigital: _&ast;nit&ast;_
I'm not a native speaker, but maybe add "the" before prompt message?
...should be downloaded when you select "yes" on **the** prompt message.
Same for the other docs.

src/extension_utils/utils.ts
```diff
@@ -302,18 +300,24 @@ export const getTelemetryState = () => {
...
+    // only use requirements file with pywin32 if on windows
+    const requirementsFileName = (process.platform === "win32") ? "requirements-windows.txt" : "requirements.txt";
+    const requirementsPath: string = getPathToScript(context, "out", requirementsFileName);
+    
+    // run command to download dependencies to out/python_libs
     const { stdout } = await exec(`${pythonExecutable} -m pip install -r ${requirementsPath} -t ${pathToLibs}`);
-    console.info(stdout);
     installed = true;
+    
+    console.log(stdout)
```
nasadigital: Maybe add a semicolon at the end of the line?
Same for error logging a few lines below.

src/extension_utils/utils.ts
```diff
@@ -302,18 +300,24 @@ export const getTelemetryState = () => {
...
-export const installPythonDependencies = async (context: vscode.ExtensionContext, pythonExecutable: string) => {
+export const installPythonDependencies = async (context: vscode.ExtensionContext, pythonExecutable: string, pathToLibs: string) => {
   let installed: boolean = false;
   try {
     vscode.window.showInformationMessage(CONSTANTS.INFO.INSTALLING_PYTHON_DEPENDENCIES);
-    const requirementsPath: string = getPathToScript(context, "out", "requirements.txt");
-    const pathToLibs: string = getPathToScript(context, "out", "python_libs");
+
+    // only use requirements file with pywin32 if on windows
+    const requirementsFileName = (process.platform === "win32") ? "requirements-windows.txt" : "requirements.txt";
```
nasadigital: Do you think it would be possible to have only windows specific requirements in `requirements-windows.txt` and all common ones in `requirements.txt`? That way a developer adding/editing/removing a requirement won't need to remember to update both files.
What do you think?

src/extension_utils/utils.ts
```diff
@@ -258,37 +258,35 @@ export const checkConfig = (configName: string): boolean => {
...
     if (checkConfig(CONFIG.SHOW_DEPENDENCY_INSTALL)) {
-      hasInstalledDependencies = await promptInstallPythonDependencies(context, pythonExecutable);
-      if (hasInstalledDependencies) {
-        await vscode.workspace
-          .getConfiguration()
-          .update(CONFIG.SHOW_DEPENDENCY_INSTALL, false);
-      }
+      // check if ./out/python_libs exists; if not, the dependencies 
+      // for adafruit_circuitpython are not (successfully) installed yet
+      hasInstalledDependencies = fs.existsSync(pathToLibs) || await promptInstallPythonDependencies(context, pythonExecutable, pathToLibs);
```
nasadigital, line 267: This is part is a bit tricky. The current check for installed dependencies is probably going to work most of the time.
Do you know what would happen if we changed requirements (is this a likely scenario?) or if a user manually removed some of the dependencies in `./out/python_libs`?

nasadigital (commented):
Solid work, very neat solution :)

src/requirements-windows.txt
andreamah: pywin32 can't download version 224 on Python 3.8 :(

README.md
andreamah: Yup! Thanks for catching that!

src/extension_utils/utils.ts
andreamah: Ooops, thanks for pointing that out!

README.md
andreamah: That sounds good! I'll make that change.

src/extension_utils/utils.ts
andreamah: That sounds good! I'll implement that.

src/requirements-windows.txt
nasadigital: Thanks for clarifying!
What about fixing it to a newer version? Are new versions compatible with Python 3.7?

src/extension_utils/utils.ts
andreamah: Ideally, the project would be re-compiled when requirements are changed. Currently, the re-compilation process doesn't delete ./out/python_libs, but I can just remove that line.

It is a sorta cumbersome process to compile every time, but I'm sure that updating requirements won't happen all the time? What do you think?

src/requirements-windows.txt
isadorasophia: I am not super familiar of what is the best practice for dealing with python dependencies, but it seems like we can do some kind of conditional dependencies for requirements.txt, [check reference](https://stackoverflow.com/a/54281345). Maybe we can try something like that instead?
It seems easier than having to maintain two different requirements files (+ we might want to add something later on for Mac?).

src/extension_utils/utils.ts
```diff
@@ -258,37 +258,35 @@ export const checkConfig = (configName: string): boolean => {
 
 export const checkPythonDependencies = async (context: vscode.ExtensionContext, pythonExecutable: string) => {
   let hasInstalledDependencies: boolean = false;
+  const pathToLibs: string = getPathToScript(context, "out", "python_libs");
```
isadorasophia: It looks like we are using the getPathToScript(context, "out", ...) in a lot of different places. Maybe we can at least define the "out" path somewhere? (E.g. let's say we decide to change the output directory, we would have to change all of its occurrences)

src/extension_utils/utils.ts
nasadigital: Thanks for clarifying. It sounds less of an issue than I thought at first. I would assume requirements wouldn't change frequently, if at all.
Up to you!

isadorasophia: Maybe you can try testing this out? E.g. what happens if we delete just one of the dependencies from `./out/python_libs`. I think the worst case scenario for this is the user having a "corrupt" environment and not being able to diagnose why.

If I understood correctly, if we delete this line we'll always prompt the user to install dependencies -- is that correct?

README.md
isadorasophia: This is not on the scope of this PR, so don't worry! I wonder if we should move these install instructions from the `README.md` and just point to the `install.md` instead? Otherwise we just have dupe information (or we can make the whole README overwhelming for new users).

src/constants.ts
```diff
@@ -45,6 +45,8 @@ export const CONSTANTS = {
       "error.debuggingSessionInProgress",
       "[ERROR] A debugging session is currently in progress, please stop it before running your code. \n"
     ),
+    DEPENDENCY_DOWNLOAD_ERROR:
+      "Package downloads failed. Some functionality may not work. ",
```
isadorasophia: Maybe we can hint the user how to address this? Something like: "Try restarting the simulator or check out [link to project install.md]".

isadorasophia (changes requested):
Nice job! ⚡️

src/extension_utils/utils.ts
andreamah: I believe that you're correct- the user would have a corrupt environment. In order to fix this, they can just re-compile.

src/requirements-windows.txt
andreamah: Added the conditional dependency! Nice catch :)

src/extension_utils/utils.ts
andreamah: I'll make "out" and "python_libs" constants in constants.ts :)

README.md
andreamah: I think that the instructions were here beforehand too (or at least the pip install commands). But yeah, it would maybe be a good idea to move all installation-related stuff to install.md for clarity (I noticed that lots of parts repeat between docs)

src/constants.ts
andreamah: Great idea! Implementing that now

src/requirements-windows.txt
andreamah: I tested pywin32 ver 227 on python 3.7 and it seemed to work! I'll fix it to ver 227.

nasadigital (approved):
Thanks for applying fixes!
LGTM :)

src/constants.ts
```diff
@@ -45,6 +45,8 @@ export const CONSTANTS = {
       "error.debuggingSessionInProgress",
       "[ERROR] A debugging session is currently in progress, please stop it before running your code. \n"
     ),
+    DEPENDENCY_DOWNLOAD_ERROR:
+      "Package downloads failed. Some functionality may not work. Try restarting the simulator or review the installation docs.",
```
vandyliu: Does "functionalities" sound better than "functionality"?

src/extension_utils/utils.ts
```diff
@@ -302,17 +301,30 @@ export const getTelemetryState = () => {
...
     vscode.window.showInformationMessage(CONSTANTS.INFO.INSTALLING_PYTHON_DEPENDENCIES);
-    const requirementsPath: string = getPathToScript(context, "out", "requirements.txt");
-    const pathToLibs: string = getPathToScript(context, "out", "python_libs");
+    
+    const requirementsPath: string = getPathToScript(context, CONSTANTS.FILESYSTEM.OUTPUT_DIRECTORY, "requirements.txt");
+
+    // run command to download dependencies to out/python_libs
     const { stdout } = await exec(`${pythonExecutable} -m pip install -r ${requirementsPath} -t ${pathToLibs}`);
-    console.info(stdout);
+    console.log(stdout);
```
vandyliu: Just curious as to why this was changed

README.md
```diff
@@ -43,19 +43,18 @@ You will be prompted to install the Python dependencies during the first use.
...
-   install by typing the following commands in a console: `pip install python-socketio`
-- _**Requests**_  
-   install by typing the following commands in a console: `pip install requests`
-- _**Application Insights**_  
-  install by typing the following commands in a console: `pip install applicationinsights`
+- Python Modules for Deploying to Device
+    - If you're on Windows, use the following command in the console to install pywin32: `pip install pywin32`
+
+- Python Modules for Simulation
+  - **Note:** On extension activation, you will be prompted with a popup message asking if you want the modules to be automatically installed for you. The following Python modules should be downloaded when you select "yes" on the prompt message. **If modules are not installed correctly, please use the "pip install" commands listed below.**
```
vandyliu, line 47: should this note be at the top/or placed somewhere else because it also applies to "Deploying to Device"

README.md
isadorasophia: Yup -- you just followed the pattern of what we have before, so it's all good! However, I think it might be worth adding a work item to do that in the future? (basically linking stuff at `README` so it's more doc-based).

README.md
isadorasophia: Just a suggestion on how we can address this, instead of having these two topics, we can have something along these lines:
(I guess the same goes for the text on Install.md and developers-setup.md)

> - Python Modules for Deploying to Device
> - Python Modules for Simulation

We can just have:
> - **(Notes here)**
>   - Python Modules
>       - Pywin32 (only windows etc)
> **Running the simulator:**
>       - (list extra modules here)

src/extension_utils/utils.ts
isadorasophia: That should be all right for now! But it might be worth adding an issue so we can handle this better in the future? If people end up hitting this particular scenario they will be able to upvote from there.

isadorasophia (approved):

src/extension_utils/utils.ts
andreamah: Oops, I'll change it back to console.info!

src/extension_utils/utils.ts
andreamah: Yeah, that'd be a good idea! I'll create an issue for it.

src/constants.ts
andreamah: yup! I'll fix that.

README.md
andreamah: I tried to make it clearer with the newest commit. Is it a bit more organized now?

markAtMicrosoft (approved):

PR #175, opened 2020-01-22
Users/t anmah/pylint import error fix

andreamah:
# Description:

Making sure that pylint doesn't flag the adafruit_circuitplayground library import as not found. This is done by suppressing the import-error in pylint for the workspace. Since pylint is often used in vscode (and users of the Python library are prompted to install it), we believed that this would be worth it.

## Type of change

# Limitations:

Configuration is not updated until extension is opened for the first time (after compilation) 

# Testing:
- [ ] Testing with/without existing configurations in settings.json

# Checklist:

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -67,6 +67,10 @@ export async function activate(context: vscode.ExtensionContext) {
 
   // Add our library path to settings.json for autocomplete functionality
   updatePythonExtraPaths();
+  
+  // ignore import errors so that adafruit_circuitpython library
+  // doesn't trigger lint errors
+  updatePylintArgs();
 
   pythonExecutableName = await utils.setPythonExectuableName();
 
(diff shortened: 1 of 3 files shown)


vandyliu:
In the readme and how-to-use.md, there is 
```
- If you have pylint enabled, it might underline the import of the adafruit_circuitplayground library, but it will work correctly.
```
Can we remove that in this PR?

src/extension.ts
```diff
@@ -880,21 +884,30 @@ const checkForTelemetry = (sensorState: any) => {
...
-  const pathToLib: string = __dirname;
-  const currentExtraPaths: string[] =
-    vscode.workspace.getConfiguration().get("python.autoComplete.extraPaths") ||
-    [];
-  if (!currentExtraPaths.includes(pathToLib)) {
-    currentExtraPaths.push(pathToLib);
+  updateConfigLists("python.autoComplete.extraPaths",[__dirname], vscode.ConfigurationTarget.Global)
+};
+
+const updatePylintArgs = () => {
```
isadorasophia, line 890: As we talked offline, I think it's worth investigating whether we can enable the adafruit import through pylint, so we don't have to disable these warnings. It looks like there is a [init-hook](https://stackoverflow.com/a/44732848) workaround in order to append a new path?

src/extension.ts
```diff
@@ -880,21 +884,30 @@ const checkForTelemetry = (sensorState: any) => {
...
+  updateConfigLists("python.linting.pylintArgs",["--disable=E0401"], vscode.ConfigurationTarget.Workspace)
```
isadorasophia, line 891: What happens if pylint is not installed?

src/extension.ts
```diff
@@ -880,21 +884,30 @@ const checkForTelemetry = (sensorState: any) => {
...
+}
+
+const updateConfigLists = (section: string, newItems: string[], scope: vscode.ConfigurationTarget) => {
+  // function for adding elements to configuration arrays
+  const currentExtraItems: string[] = vscode.workspace.getConfiguration().get(section) || [];
+
+  for (const item of newItems) {
+    if (!currentExtraItems.includes(item)) {
```
isadorasophia, line 899: You can also use a python `set()` instead

isadorasophia (changes requested):
that seems like a very confusing issue for inexperienced devs, nice job handling it! 🌟

src/extension.ts
andreamah: I looked into init-hook and am trying to debug the issues :0. Looking into it, I believe that the previous group tried to add the python_libs to the system path already in process_user_code.py (~lines 15-18). However, when checking the included modules in pylint (using `pylint --init-hook 'import sys; print(sys.modules.keys())'`) , the libraries aren't there :(. I'll try to look more into it!

andreamah:
Seems a bit too high-effort for the importance. Will re-do later if necessary.

PR #180, opened 2020-01-25
WIP: shim design example

andreamah:
DON'T MERGE; just an small-scale example of the python microbit model design (as requested by coaches)


andreamah:
@sagarmanchanda is this what you envisioned in terms of the shim-based singleton design?

sagarmanchanda:
> @sagarmanchanda is this what you envisioned in terms of the shim-based singleton design?

sagarmanchanda (commented):
Looks good. This is exactly what I was thinking. 

Although we should now think about the folder structuring too.

src
|____microbit
_____|____tests
_____|____microbit_model
__________|____microbit_model.py
__________|____display # Could be a folder if it has multiple files, otherwise just files in root
_____|____shim.py
_____|____init.py
       
What do you guys think about the folder structuring?

andreamah:
@sagarmanchanda I think that that directory structure sounds good! I'll try to incorporate that into the upcoming work.

Also, I was thinking that it might be better to replace 

```
class display:
    @staticmethod
    def scroll(self, message):
        microbit_model.mb.display.scroll(self,message) 
```
in src/microbit/code_processing_shim.py with 

`display = microbit_model.mb.display`

To reduce repetition with static methods mapping to actual object methods. I believe that the new version uses pass-by-reference in order to assign a simple global "display" object with the microbit model's display.

What do you think?

src/microbit/code_processing_shim.py
```diff
@@ -0,0 +1,13 @@
+from . import microbit_model
+
+# EXAMPLE
+# can be called simply as "show_message("string")"
+def show_message(message):
```
xnkevinnguyen, line 7: That's awesome, it was exactly what I was thinking as well :)

sagarmanchanda:
@andreamah Good catch! Totally makes sense to directly use the methods rather than creating static mappings.

sagarmanchanda:
Also thanks for doing this sample implementation! Good job. :)

andreamah:
no intention to merge these changes yet, will ping coaches with new changes when it comes up

PR #181, opened 2020-01-25
Users/t anmah/pylint import error fix

andreamah:
# Description:
Update: I took another look at the pylint import error and I think I was able to get init-hook to recognize the libraries that are local to the extention!! Originally, I couldn't figure it out and closed the original PR.

Original Issue: All references to the local libraries (like adafruit_circuitplayground) were flagged as missing imports by pylint.

## Type of change

# Limitations:

# Testing:

- [ ] Test with importing local libraries + checks with non-installed libraries to see if import errors still come up

# Checklist:


src/extension.ts
```diff
@@ -973,23 +977,51 @@ const checkForTelemetry = (sensorState: any) => {
...
+                        `import sys; sys.path.extend([\"${outPath}\",\"${pyLibsPath}\"])`
+                    ], 
+                    vscode.ConfigurationTarget.Workspace)
+}
+
+const createEscapedPath = (pieces: string[]) => {
+  const initialPath:string = pieces.join("\\");
+
+  // escape all instances of backslashes
+  return initialPath.replace(/\\/g,"\\\\");
```
isadorasophia: You can also use:
 
```python
import re
re.escape(str)
```

Assuming we want to escape for regex?

isadorasophia (approved):
yay :D

src/extension.ts
andreamah: I tried to use escape(), but it also escaped other characters (like semicolons) :(

andreamah: ^ oh wait, your response is Python (I think). I'm using Typescript.

src/extension.ts
```diff
@@ -973,23 +977,51 @@ const checkForTelemetry = (sensorState: any) => {
...
-        currentExtraPaths.push(pathToLib);
-    }
-    vscode.workspace
-        .getConfiguration()
-        .update(
-            "python.autoComplete.extraPaths",
-            currentExtraPaths,
-            vscode.ConfigurationTarget.Global
-        );
+  updateConfigLists("python.autoComplete.extraPaths",[__dirname], vscode.ConfigurationTarget.Global)
```
vandyliu: run prettier formatting 🥺

vandyliu (changes requested):
run prettier & should be good 💃

src/extension.ts
andreamah: done!

src/extension.ts
isadorasophia: We talked offline, we don't really need to escape this for regex, but rather just the backlashes.

isadorasophia (approved):

vandyliu (approved):

PR #185, opened 2020-01-30
Creating the skeleton of the microbit library and adding a button object

vandyliu:
# Description:

Adding the skeleton/structure of the microbit library. This was based off Andrea's closed [PR](https://github.com/microsoft/vscode-python-devicesimulator/pull/180), which Sagar commented on. Also included the basic button class in it.

## Type of change

# Limitations:

# Testing:

- [ ] Test A
- [ ] Test B

# Checklist:


src/microbit/model/microbit.py
```diff
@@ -0,0 +1,15 @@
...
+
+class MicrobitModel:
+    def __init__(self):
+        # State in the Python process
+        self.button_a = Button()
+        self.button_b = Button()
+
+    # Will be removed once functions added to shim
+    def show_message(self, message):
+        print("message: " + message)
```
xnkevinnguyen: nit: should have To be implemented, also it seems this function is already called in the shim

vandyliu: done, added real functions that are part of microbit library

src/microbit/model/button.py
```diff
@@ -0,0 +1,26 @@
...
+
+    def was_pressed(self):
+        res = self.__prev_pressed
+        self.__prev_pressed = False
+        return res
+
+    def get_presses(self):
+        res = self.__presses
+        self.__presses = 0
+        return res
```
nasadigital, line 19: These two methods are kind of weird in that they change the state while checking for the value. Just curious, when you are implementing these classes is there any code you can reference?

src/microbit/shim.py
```diff
@@ -0,0 +1,9 @@
+from .model import microbit_model
+
+
+def sleep(n):
+    microbit_model.mb.sleep(n)
+
+
+def running_time():
+    microbit_model.mb.running_time()
```
nasadigital, line 9: _&ast;nit&ast;_
I think there might be an extra newline between these functions.

gulpfile.js
```diff
@@ -1,135 +1,137 @@
...
+      "../../dist/*0.0.0-UNTRACKEDVERSION.vsix"
+    ],
+    { force: true }
+  );
+});
+
+const pythonToMove = [
+  "./src/adafruit_circuitplayground/*.*",
+  "./src/microbit/*.*",
+  "./src/microbit/model/*.*",
```
nasadigital: Do you know if `"./src/microbit/*.*"` includes the model directory?

gulpfile.js
```diff
@@ -1,135 +1,137 @@
...
+      "package.nls.*.json",
+      "../../dist/*0.0.0-UNTRACKEDVERSION.vsix"
+    ],
+    { force: true }
+  );
+});
+
+const pythonToMove = [
+  "./src/adafruit_circuitplayground/*.*",
+  "./src/microbit/*.*",
```
andreamah: Maybe it would be a good idea to list one path that encapsulates everything recursively inside ./src/microbit/ except ./src/microbit/tests/*.*. Try "./src/microbit/!(test)/*.*"

andreamah (changes requested):
Good! Awesome work :)

src/microbit/test/test_button.py
```diff
@@ -0,0 +1,41 @@
...
+        assert was_pressed == self.button.was_pressed()
+        # Button resets prev pressed after was_pressed() is called
+        assert not self.button.was_pressed()
+
+    @pytest.mark.parametrize("presses", [0, 2, 4])
+    def test_get_presses(self, presses):
+        self.button._Button__presses = presses
+        assert presses == self.button.get_presses()
+        # Presses is reset to 0 after get_presses() is called
+        assert 0 == self.button.get_presses()
```
nasadigital: If my understanding is correct, the expected usage of the Button class is through its methods and not by directly modifying its state.
If we don't expect to use this class by modifying its state directly, I think we can modify/remove these three tests since they will slow us down when we refactor the Button class.
In my opinion, we would get most value out by modifying the tests to use the implemented methods (`__press_down` and `__release`) like in the next test (`test_press_down`).
Let me know what you think and if that makes sense :)

src/microbit/model/button.py
```diff
@@ -0,0 +1,26 @@
...
+
+    def __press_down(self):
```
nasadigital, line 21: How do we want to handle if this function is called twice in a row? Do we want to register a release between the presses? Should __prev_pressed be set to true?

gulpfile.js
vandyliu: done!

src/microbit/test/test_microbit_model.py
```diff
@@ -0,0 +1,24 @@
...
+    @pytest.mark.parametrize("value", [9, 30, 1999])
+    def test_sleep(self, value):
+        time.sleep = mock.Mock()
+        self.mb.sleep(value)
+        time.sleep.assert_called_with(value / 1000)
+
+    def test_running_time(self):
+        mock_start_time = 10
+        mock_end_time = 300
+        self.mb._MicrobitModel__start_time = mock_start_time
```
nasadigital, line 21: Just FYI, (you might already be familiar with this) you might have the constructor be something like:
```
class MicrobitModel:
    def __init__(self, start_time=time.time()):
        # State in the Python process
        self.button_a = Button()
        self.button_b = Button()
        self.__start_time = start_time
```
By exposing the "hidden" parameter, it is easier to test that functionality of the class.
Because we want to be 1:1 with the micro:bit library, we can't do that this time though.

src/microbit/test/test_shim.py
```diff
@@ -0,0 +1,19 @@
...
+
+import pytest
+from unittest import mock
+from .. import shim
+from ..model import microbit_model
+
+
+class TestShim(object):
+    def test_sleep(self):
+        n = 100
```
nasadigital: _&ast;nit&ast;_
Maybe name this variable something more descriptive?

nasadigital (commented):
I am impressed how thorough the tests are! Great job.

src/microbit/shim.py
nasadigital: It seems that this is the code style you are using. Disregard this comment :)

src/microbit/model/button.py
vandyliu: https://github.com/bbcmicrobit/micropython/blob/master/docs/button.rst

I'm referring to the micropython docs. I will add these docs to the top to make it more clear as to why the implementation is like that though.

src/microbit/test/test_microbit_model.py
vandyliu: ooo, i did not know that. Thank you :D
I won't be changing it but it is nice to know :)

src/microbit/test/test_shim.py
vandyliu: fixed

src/microbit/model/button.py
nasadigital: These docs are much more extensive than I thought. Thanks for sharing!

src/microbit/model/button.py
vandyliu: The front end side will be calling these functions, and so it can register clicks and releases.
If someone is holding it, then I wouldn't want it to call release right away. When that person releases their button click, it will call the release function. Theoretically, __press_down should not be called twice without a release between them. Let me know what you think or if there might be any problems with that logic.

src/microbit/test/test_button.py
vandyliu: I updated the tests. Let me know what you think. I originally had it that way because so it wouldn't rely on __press_down and __release in case those functions were broken.

gulpfile.js
vandyliu: it does not

src/microbit/model/button.py
nasadigital: Up to you, just double checking whether we are considering these cases.

src/microbit/model/button.py
```diff
@@ -0,0 +1,27 @@
...
+        self.__pressed = True
+        self.__presses += 1
+
+    def __release(self):
+        self.__pressed = False
+        self.__prev_pressed = True
```
nasadigital: According to the docs it might be more accurate to set `__prev_pressed` to true in `__press_down()`:
```
Returns ``True`` or ``False`` to indicate if the button was pressed
(went from up to down) since the device started or the last time this
method was called.
```
What do you think?

src/microbit/test/test_button.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def test_get_presses_two(self):
+        assert 0 == self.button.get_presses()
+        self.button._Button__press_down()
+        self.button._Button__release()
+        self.button._Button__press_down()
+        self.button._Button__release()
+        assert 2 == self.button.get_presses()
+        # Presses is reset to 0 after get_presses() is called.
+        assert 0 == self.button.get_presses()
```
nasadigital: Just FYI, you could still parametrize this function by having a for loop call `press_down()` and `release()` as many times as the parameter.

src/microbit/model/button.py
vandyliu: Def considered! Will be aware of it and thinking about the cases for the integration.

nasadigital (commented):
Thanks for addressing the comments!

src/microbit/test/test_button.py
vandyliu: You are very right! Will change this

src/microbit/model/button.py
vandyliu: I agree, will change it.

gulpfile.js
nasadigital: Thanks for confirming!

src/microbit/test/test_button.py
```diff
@@ -51,4 +51,3 @@ def test_get_presses_two(self):
         assert 2 == self.button.get_presses()
         # Presses is reset to 0 after get_presses() is called.
         assert 0 == self.button.get_presses()
-
```
nasadigital: Just FYI, but you can set up [pre-commit hooks](https://pre-commit.com/) on your machine to automatically run checks and fixes on your commits before you send them.

vandyliu: I'll look into this! Ty

nasadigital (approved):
LGTM! Great PR 🌠

andreamah (approved):
Looks good! :D

sagarmanchanda (approved):

PR #186, opened 2020-01-31
Users/t anmah/python image model initial

andreamah:
# Description:

Created initial work on Image class. The remaining work just deals with having Image.HEART or any other Image.[CONSTANT] return a new read-only Image object. 

In this PR, the Image class should have all functions for manipulation.

## Type of change

- [ x] New feature (non-breaking change which adds functionality)\

# Limitations:

Image class not completely finished.

# Testing:

- [ ] src/microbit/test/test_image.py

# Checklist:


src/microbit/model/constants.py
```diff
@@ -0,0 +1,19 @@
+INDEX_ERR = "index out of bounds"
+BRIGHTNESS_ERR = "brightness out of bounds"
+SAME_SIZE_ERR = "images must be the same size"
+UNSUPPORTED_ADD_TYPE = "unsupported types for __add__:"
+INCORR_IMAGE_SIZE = "image data is incorrect size"
+LED_WIDTH = 5
```
xnkevinnguyen: nit: group your constants by categories

src/microbit/model/image.py
```diff
@@ -0,0 +1,268 @@
...
+
+            if len(args) == 3:
+                # This option is for potential third bytearray arguments
+                byte_arr = args[2]
+                self.__LED = self.__bytes_to_array(width, height, byte_arr)
+            else:
+                self.__LED = self.__create_leds(width, height)
+
+    def width(self):
+        if len(self.__LED):
```
xnkevinnguyen: since you check if self.__LED exists, shouldn't it be `if self.__LED`  ?

andreamah: this check sees if self.__LED is empty. This must be checked before we try to access its first element.

src/microbit/model/constants.py
andreamah: ok I'll make that change!

src/microbit/model/image.py
nasadigital: Just FYI:
```
test_list = []
if test_list:
    print("This list has elements")
else:
    print("This list is empty")
```
The above code will print `This list is empty`.
I prefer the current check `len(self.__LED)`, but it's up to you.

xnkevinnguyen: I see! Didn't notice you were accessing the first element :)

src/microbit/model/constants.py
```diff
@@ -0,0 +1,16 @@
+BLANK = "00000:00000:00000:00000:00000:"
+BOAT = "05050:05050:05050:99999:09990:"
+BRIGHTNESS_ERR = "brightness out of bounds"
+COPY_ERR_MESSAGE = "please copy() first"
```
nasadigital: Are we considering internationalization?
Hard coding these messages, would make i18n a bit difficult in the future.

src/microbit/model/constants.py
```diff
@@ -0,0 +1,16 @@
+BLANK = "00000:00000:00000:00000:00000:"
+BOAT = "05050:05050:05050:99999:09990:"
+BRIGHTNESS_ERR = "brightness out of bounds"
+COPY_ERR_MESSAGE = "please copy() first"
+
+HEART = "09090:99999:99999:09990:00900:"
+INCORR_IMAGE_SIZE = "image data is incorrect size"
+INDEX_ERR = "index out of bounds"
```
vandyliu: have all the error messages in one section

Kind of like:
```
# Errors
INCORR_IMAGE_SIZE = "image data is incorrect size"
INDEX_ERR = "index out of bounds"
etc
```

src/microbit/model/constants.py
andreamah: We are likely going to deal with optimizing for localization later, since we're not sure if we'll have time for it.

src/microbit/model/image.py
```diff
@@ -0,0 +1,289 @@
+# implementing image model as described here:
+# https://microbit-micropython.readthedocs.io/en/latest/image.html
```
vandyliu: Noticed the placement of this is different from Button class. We should probably keep it consistent.

src/microbit/model/constants.py
```diff
@@ -0,0 +1,16 @@
+BLANK = "00000:00000:00000:00000:00000:"
```
vandyliu: Nit: Maybe be more descriptive with this. 
like BLANK_5X5

src/microbit/model/image.py
```diff
@@ -0,0 +1,289 @@
...
+    def set_pixel(self, x, y, value):
+        try:
+            if not self.__valid_pos(x, y):
+                raise ValueError(CONSTANTS.INDEX_ERR)
+            elif not self.__valid_brightness(value):
+                raise ValueError(CONSTANTS.BRIGHTNESS_ERR)
+            else:
+                self.__LED[y][x] = value
+        except TypeError:
+            print(CONSTANTS.COPY_ERR_MESSAGE)
```
vandyliu: For this function, is it correct to be printing the message?

src/microbit/model/image.py
```diff
@@ -0,0 +1,289 @@
...
+    def crop(self, x, y, w, h):
+        res = Image(w, h)
+        res.blit(self, x, y, w, h)
+        return res
+
+    def copy(self):
+        return Image(self.__create_string())
+
+    # This inverts the brightness of each LED.
+    # ie: pixel that is at brightness 4 would become brightness 5
```
vandyliu: nit: capitlize pixel

src/microbit/model/image.py
```diff
@@ -0,0 +1,289 @@
...
+
+    # This fills all LEDs with same brightness.
+    def fill(self, value):
+        for y in range(0, self.height()):
+            for x in range(0, self.width()):
+                self.set_pixel(x, y, value)
+
+    # This transposes certain area (w x h) on src onto current image.
+    def blit(self, src, x, y, w, h, xdest=0, ydest=0):
+
```
vandyliu: nit: remove space

src/microbit/model/image.py
```diff
@@ -0,0 +1,289 @@
...
+            for x in range(0, self.width()):
+                self.set_pixel(x, y, 9 - self.get_pixel(x, y))
+
+    # This fills all LEDs with same brightness.
+    def fill(self, value):
+        for y in range(0, self.height()):
+            for x in range(0, self.width()):
+                self.set_pixel(x, y, value)
+
+    # This transposes certain area (w x h) on src onto current image.
```
vandyliu: Nit: revise the sentence
```
# This transposes certain area (w x h) on src onto current image.
```
turn to:
```
# This transposes a certain area (w x h) of src onto the current image.
```

src/microbit/model/image.py
```diff
@@ -0,0 +1,289 @@
...
+
+        return res
+
+    # HELPER FUNCTIONS
+
+    # This create 2D array of off LEDs with
+    # width w and height h
+    def __create_leds(self, w, h):
+        arr = []
+        for _ in range(0, h):
```
vandyliu: nit: just make range(h)
same for below

src/microbit/model/image.py
```diff
@@ -0,0 +1,289 @@
...
+                sub_arr.append(0)
+            arr.append(sub_arr)
+
+        return arr
+
+    # This turns byte array to 2D array for LED field.
+    def __bytes_to_array(self, width, height, byte_arr):
+        bytes_translated = bytes(byte_arr)
+
+        if not (len(bytes_translated)) == height * width:
```
vandyliu: nit: remove brackets outside of len

src/microbit/model/image.py
```diff
@@ -0,0 +1,289 @@
...
+            return result
+
+    def __valid_brightness(self, value):
+        return value >= 0 and value <= 9
+
+    def __valid_pos(self, x, y):
+        return x >= 0 and x < self.width() and y >= 0 and y < self.height()
+
+    def __shift_vertical(self, n):
+
```
vandyliu: nit; remove this space

src/microbit/test/test_image.py
```diff
@@ -0,0 +1,294 @@
+import pytest
+from ..model.image import Image
+
+from ..model import constants as CONSTANTS
+
+# TESTING FOR IMAGE CLASS
```
vandyliu: don't need this comment

src/microbit/test/test_image.py
```diff
@@ -0,0 +1,294 @@
...
+
+# TESTING FOR IMAGE CLASS
+
+
+class TestImage(object):
+    def setup_method(self):
+        self.image = Image()
+        self.image_heart = Image(CONSTANTS.HEART)
+
+    # GET PIXEL
```
vandyliu: Don't think these comments are necessary because the function says this

src/microbit/model/constants.py
andreamah: corrected in next commit!

src/microbit/model/image.py
```diff
@@ -0,0 +1,289 @@
...
+        for arr_y in initial_array:
+            num_extra_spaces = max_subarray_len - len(arr_y)
+            for _ in range(num_extra_spaces):
+                arr_y.append(0)
+
+        return initial_array
+
+    def __string_to_initial_array(self, pattern):
+        # The result may have spaces in the 2D array
+        # and may unevent sub-array lengths
```
vandyliu: what does unevent mean here?

src/microbit/model/constants.py
```diff
@@ -0,0 +1,16 @@
+BLANK = "00000:00000:00000:00000:00000:"
+BOAT = "05050:05050:05050:99999:09990:"
```
vandyliu, line 3: Also maybe we should have another file for all these image constants because this file will get cluttered very quickly with all the images.

src/microbit/model/image.py
andreamah: oops, will fix ^

src/microbit/model/constants.py
andreamah: For now we are keeping these here since we don't know how we're going to access all of the shapes (it can be like the letters where they are parsed from a larger array).

src/microbit/model/image.py
andreamah: nope, I'll fix it in the next commit!

src/microbit/test/test_image.py
```diff
@@ -0,0 +1,294 @@
...
+            (0, -1, 0, CONSTANTS.INDEX_ERR),
+            (0, 0, -1, CONSTANTS.BRIGHTNESS_ERR),
+        ],
+    )
+    def test_set_pixel_error(self, x, y, brightness, err_msg):
+        with pytest.raises(ValueError, match=err_msg):
+            self.image.set_pixel(x, y, brightness)
+
+    # WIDTH & HEIGHT
+    @pytest.mark.parametrize("image", [(Image()), (Image(3, 3)), (Image(""))])
```
vandyliu: maybe also add tests for non-square images

src/microbit/model/image.py
```diff
@@ -0,0 +1,289 @@
...
+        arr = []
+        sub_arr = []
+
+        max_subarray_len = 0
+
+        for elem in pattern:
+            if elem == ":":
```
vandyliu: Awesome stuff. We should also add the ability to separate with '\n' as seen here:
https://github.com/bbcmicrobit/micropython/blob/master/docs/image.rst

src/microbit/model/image.py
```diff
@@ -0,0 +1,289 @@
...
+        for index_y in range(0, self.height()):
+            ret_str += self.__row_to_str(index_y)
+        return ret_str
+
+    def __row_to_str(self, y):
+        new_str = ""
+        for x in range(0, self.width()):
+            new_str = new_str + str(self.get_pixel(x, y))
+
+        new_str = new_str + ":"
```
vandyliu: nit: keep it consistent with the __create_string function and use new_str += ______________ or new_str = new_str + ____

vandyliu: same for in the foreach loop above and functions below

vandyliu (changes requested):
Absolutely awesome work!!! Excited to merge this in!!!!!!!!!!!!!!!!!!!

src/microbit/test/test_image.py
```diff
@@ -0,0 +1,278 @@
...
+    # @pytest.mark.parametrize(
+    #     "initial_img_string, actual",
+    #     [
+    #         ("0:000:00:0000:", Image("0000:0000:0000:0000:")),
+    #         ("0000\n0000\n0000\n0000\n", Image("0000:0000:0000:0000:")),
+    #         ("0\n000\n00\n0000\n", Image("0000:0000:0000:0000:")),
+    #     ],
+    # )
+    # def test_string_processing_constructor(self, initial_img_string, actual):
+    #     assert Image(initial_img_string)._Image__LED == actual._Image__LED
```
nasadigital: Perhaps remove this comment. If we want to include this code in the future, the author can find it on git.

src/microbit/model/image.py
```diff
@@ -0,0 +1,290 @@
...
+
+    def set_pixel(self, x, y, value):
+        try:
+            if not self.__valid_pos(x, y):
+                raise ValueError(CONSTANTS.INDEX_ERR)
+            elif not self.__valid_brightness(value):
+                raise ValueError(CONSTANTS.BRIGHTNESS_ERR)
+            else:
+                self.__LED[y][x] = value
+        except TypeError:
```
nasadigital: When do we expect this TypeError to be raised?
Is it in: `self.__LED[y][x] = value` ?

src/microbit/model/image.py
```diff
@@ -0,0 +1,290 @@
...
+        return self.__shift_vertical(n * -1)
+
+    def shift_down(self, n):
+        return self.__shift_vertical(n)
+
+    def shift_right(self, n):
+        return self.__shift_horizontal(n)
+
+    def shift_left(self, n):
+        return self.__shift_horizontal(n * -1)
```
nasadigital: Perhaps use `-n` rather than `n * -1` in these functions?

src/microbit/model/image.py
```diff
@@ -0,0 +1,290 @@
...
+            raise ValueError(CONSTANTS.INDEX_ERR)
+
+        for count_y in range(0, h):
+            for count_x in range(0, w):
+                if self.__valid_pos(xdest + count_x, ydest + count_y):
+                    if src.__valid_pos(x + count_x, y + count_y):
+                        transfer_pixel = src.get_pixel(x + count_x, y + count_y)
+                    else:
+                        transfer_pixel = 0
+                    self.set_pixel(xdest + count_x, ydest + count_y, transfer_pixel)
```
nasadigital: I'm not confident this method would work as intended if self and src are the same object.
What would be the result for the following example:
```
self = src = [[1, 2, 3], [4, 5, 6], [7, 8, 9]]
x, y = 0, 0
w, h = 2, 2
xdest, ydest = 1, 1
```

src/microbit/model/image.py
```diff
@@ -0,0 +1,290 @@
...
+        return res
+
+    def copy(self):
+        return Image(self.__create_string())
+
+    # This inverts the brightness of each LED.
+    # ie: Pixel that is at brightness 4 would become brightness 5
+    # and pixel that is at brightness 9 would become brightness 0.
+    def invert(self):
+        for y in range(0, self.height()):
```
nasadigital: I would omit the first argument when it's zero.

src/microbit/model/image.py
```diff
@@ -0,0 +1,290 @@
...
+
+        # Fill in empty spaces in w x h matrix.
+        for arr_y in initial_array:
+            num_extra_spaces = max_subarray_len - len(arr_y)
+            for _ in range(num_extra_spaces):
+                arr_y.append(0)
+
+        return initial_array
+
+    def __string_to_initial_array(self, pattern):
```
nasadigital: These method names are a bit confusing, I have to look at the code to see what's the difference between `__string_to_array` and `__string_to_initial_array`.

src/microbit/model/image.py
```diff
@@ -0,0 +1,290 @@
...
+        # Depending on the number of arguments
+        # in constructor, it treat args differently.
+
+        if len(args) == 0:
+            # default constructor
+            self.__LED = self.__string_to_array(CONSTANTS.BLANK_5X5)
+        elif len(args) == 1:
+            pattern = args[0]
+            if isinstance(pattern, str):
+                self.__LED = self.__string_to_array(pattern)
```
nasadigital: Are there any checks for the pattern dimensions?

nasadigital (commented):
Great job, there was a lot to implement! I put in a few comments.

src/microbit/model/image.py
```diff
@@ -0,0 +1,295 @@
...
+            arr.append(sub_arr)
+
+        return arr, max_subarray_len
+
+    # This returns the limit if the result is too big.
+    def __limit_result(self, limit, result):
+        if result > limit:
+            return limit
+        else:
+            return result
```
nasadigital: You could remove this method and replace it with method with `min(limit, result)`.

src/microbit/model/image.py
```diff
@@ -0,0 +1,295 @@
...
+
+    def __valid_brightness(self, value):
+        return value >= 0 and value <= 9
```
nasadigital: I would put the brightness limits in constant variables. They are used often, so it would also add clarity to most methods.

src/microbit/model/image.py
nasadigital: My bad I assumed that the height and width can be at most 5x5.

src/microbit/test/test_image.py
```diff
@@ -0,0 +1,288 @@
...
+
+    @pytest.mark.parametrize("x, y, brightness", [(1, 1, 4), (2, 3, 6), (4, 4, 9)])
+    def test_get_pixel(self, x, y, brightness):
+        self.image._Image__LED[y][x] = brightness
+        assert brightness == self.image.get_pixel(x, y)
+
+    @pytest.mark.parametrize("x, y, brightness", [(1, 1, 4), (2, 3, 6), (4, 4, 9)])
+    def test_set_pixel(self, x, y, brightness):
+        self.image.set_pixel(x, y, brightness)
+        assert brightness == self.image._Image__LED[y][x]
```
nasadigital: I would merge these two into one test and only test `set_pixel/get_pixel` instead of `__LED`.
In general we want to focus on testing behavior, rather than implementation. If we change the implementation we don't want to have tests failing.

nasadigital (commented):
A few more comments.
I really appreciate how you tried to make the tests easy to understand to the reader, even though it's difficult!

src/microbit/test/test_image.py
andreamah: oops, didn't mean to leave that in! thanks :)

src/microbit/model/image.py
andreamah: Oops, that was from before when the internal data structure was a tuple for non-editable instances. I'll remove it.

src/microbit/model/image.py
andreamah: Can I just add 

        if self == src:
            src = src.copy()

 To fix this?

nasadigital: Yup, that should do it (you might even drop the if clause, up to you). I think adding a comment regarding the motivation would probably help future contributors.
I would also add a unit test to catch this corner case.

src/microbit/model/image.py
```diff
@@ -123,10 +123,10 @@ def __add__(self, other):
         else:
             res = Image(self.width(), self.height())
 
-            for y in range(0, self.height()):
-                for x in range(0, self.width()):
+            for y in range(self.height()):
+                for x in range(self.width()):
                     sum_value = other.get_pixel(x, y) + self.get_pixel(x, y)
-                    display_result = self.__limit_result(9, sum_value)
+                    display_result = min(9, sum_value)
```
nasadigital: You can use `BRIGHTNESS_MAX` instead. Likewise in row 146.

nasadigital (approved):
Thank you for applying the fixes! LGTM 💯

vandyliu (approved):

PR #187, opened 2020-02-03
Initial Buttons Set up

xnkevinnguyen:
# Description:
Make 3 buttons clickable, current behaviour is not implemented yet.

## Type of change

# Limitations:

# Testing:

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/src/view/components/cpx/CpxSimulator.tsx
+++ b/src/view/components/cpx/CpxSimulator.tsx
@@ -2,14 +2,14 @@
 // Licensed under the MIT license.
 
 import * as React from "react";
-import { CONSTANTS } from "../constants";
-import "../styles/Simulator.css";
-import PlayLogo from "../svgs/play_svg";
-import StopLogo from "../svgs/stop_svg";
-import { BUTTON_NEUTRAL, BUTTON_PRESSED } from "./cpx/Cpx_svg_style";
-import { CpxImage, updatePinTouch, updateSwitch } from "./cpx/CpxImage";
-import Dropdown from "./Dropdown";
-import ActionBar from "./simulator/ActionBar";
+import { CONSTANTS } from "../../constants";
+import "../../styles/Simulator.css";
+import PlayLogo from "../../svgs/play_svg";
+import StopLogo from "../../svgs/stop_svg";
+import { BUTTON_NEUTRAL, BUTTON_PRESSED } from "./Cpx_svg_style";
+import { CpxImage, updatePinTouch, updateSwitch } from "./CpxImage";
+import Dropdown from "../Dropdown";
+import ActionBar from "../simulator/ActionBar";
 
 interface ICpxState {
     pixels: number[][];
--- a/src/view/components/cpx/CpxSimulator.tsx
+++ b/src/view/components/cpx/CpxSimulator.tsx
@@ -29,9 +29,6 @@ interface IState {
     cpx: ICpxState;
     play_button: boolean;
 }
-interface IProps {
-    children?: any;
-}
 
 const DEFAULT_CPX_STATE: ICpxState = {
     brightness: 1.0,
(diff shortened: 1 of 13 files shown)


nasadigital:
@sagarmanchanda What do you think of the current implementation of the Microbit component? I think something similar is used in cpx, as well. To me accessing elements by id seems like it could cause us trouble in the future.
Ideally the elements would be components themselves and we wouldn't access them through ids, but I am not sure how much effort that is.
Not super familiar with these kinds of things, so interested to hear your thoughts!

xnkevinnguyen:
> @sagarmanchanda What do you think of the current implementation of the Microbit component? I think something similar is used in cpx, as well. To me accessing elements by id seems like it could cause us trouble in the future.
> Ideally the elements would be components themselves and we wouldn't access them through ids, but I am not sure how much effort that is.
> Not super familiar with these kinds of things, so interested to hear your thoughts!

@nasadigital  The current implementation is a similar one to the cpx, one potential implementation would be to make components and attach them to the current existing svg or overlay on it. I feel the same way that making regular components would lead to higher code clarity in the future. I could start doing this with LEDS. It also makes sense in the way that I won't have to take care of the rerendering (LEDS)

nasadigital:
> @nasadigital The current implementation is a similar one to the cpx, one potential implementation would be to make components and attach them to the current existing svg or overlay on it. I feel the same way that making regular components would lead to higher code clarity in the future. I could start doing this with LEDS

@xnkevinnguyen Thanks for clarifying! I agree and also it would probably be easier for development. I am not sure if there are any other drawbacks to this approach, though...
For cpx I don't think we should make any changes if everything works.

src/view/components/microbit/MicrobitImage.tsx
```diff
@@ -0,0 +1,52 @@
...
+    buttonElement.onmousedown = e => {
+        eventTriggers.onMouseDown(buttonElement, e);
+    };
+    buttonElement.onmouseup = e => {
+        eventTriggers.onMouseUp(buttonElement, e);
+    };
+
+    buttonElement.onmouseleave = e => {
+        eventTriggers.onMouseLeave(buttonElement, e);
+    };
```
nasadigital: The spacing between the statements is incosistent.

nasadigital (approved):
LGTM!
shipit

src/view/components/cpx/CpxSimulator.tsx
```diff
@@ -29,7 +29,7 @@ interface IState {
     cpx: ICpxState;
     play_button: boolean;
 }
-interface IMyProps {
+interface IProps {
```
sagarmanchanda: I don't think you need to define children as a prop, as it already available. So even if you pass empty props here, in your component this.props.children would still be available

src/view/components/microbit/MicrobitImage.tsx
```diff
@@ -0,0 +1,52 @@
...
+    onMouseLeave: (button: HTMLElement, event: Event) => void;
+}
+interface IProps {
+    eventTriggers: EventTriggers;
+}
+
+// Displays the SVG and call necessary svg modification.
+export class MicrobitImage extends React.Component<IProps, any> {
+    componentDidMount() {
+        const svgElement = window.document.getElementById("microbit_svg");
```
sagarmanchanda: Look at how refs and findDOMNode works, there are always preferable ways than getElementById

sagarmanchanda (changes requested):
Can you please check in the first PR and rebase again, I still see a lot of repeated changes here. Thanks! :)

src/view/components/cpx/CpxSimulator.tsx
xnkevinnguyen: Yes, i'm making the change to the existing codebase

src/view/components/microbit/MicrobitImage.tsx
xnkevinnguyen: I will be using a reference as suggested, it is best practice in case of duplicates

PR #189, opened 2020-02-04
Display module for microbit library

vandyliu:
# Description:

Implemented the display module for the microbit library.

## Type of change

# Limitations:

# Testing:

- [ ] Test A
- [ ] Test B

# Checklist:


src/microbit/model/constants.py
```diff
@@ -95,6 +95,11 @@
...
+ASCII_START = 32
```
andreamah, line 102: perhaps add a comment to explain what these are?

src/microbit/model/display.py
```diff
@@ -0,0 +1,207 @@
...
+                    chars = list(str(value))
+
+                for c in chars:
+                    self.__image = Display.__get_image_from_char(c)
+                    time.sleep(delay / 1000)
+            else:
+                # Check if iterable
+                try:
+                    _ = iter(value)
+                except TypeError as e:
```
andreamah: can't we just let the typeerror get thrown (instead of catching it and re-raising it)? or do you only want this behavior when typeerrors are thrown?

src/microbit/model/display.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def is_on(self):
+        return self.__on
+
+    def read_light_level(self):
+        raise NotImplementedError(CONSTANTS.NOT_IMPLEMENTED_ERROR)
+
+    # Helpers
+
+    def __get_array(self):
```
andreamah, line 174: consider doing a deep copy?

src/microbit/model/display.py
```diff
@@ -0,0 +1,207 @@
...
+        return self.__image._Image__LED
+
+    def __print(self):
+        print("")
```
andreamah: is this "print("")" part of the spec?

src/microbit/model/display.py
```diff
@@ -0,0 +1,207 @@
...
+        for i in range(CONSTANTS.LED_HEIGHT):
+            print(self._Display__image._Image__LED[i])
+
+    @staticmethod
+    def __get_image_from_char(c):
+        # If c is not between the ASCII alphabet we cover, make it a question mark
+        if ord(c) < CONSTANTS.ASCII_START or ord(c) > CONSTANTS.ASCII_END:
+            c = "?"
+        offset = (ord(c) - CONSTANTS.ASCII_START) * 5
+        representative_bytes = CONSTANTS.ALPHABET[offset : offset + 5]
```
andreamah: add a small comment to explain this? (particularly, why 5?)

src/microbit/model/display.py
```diff
@@ -0,0 +1,207 @@
...
+        max_index = 0
+        for row in image._Image__LED:
+            for index, bit in enumerate(row):
+                if bit > 0:
+                    min_index = min(min_index, index)
+                    max_index = max(max_index, index)
+        return image.crop(min_index, 0, max_index - min_index + 1, CONSTANTS.LED_HEIGHT)
+
+    @staticmethod
+    def __convert_bytearray_to_image_str(byte_array):
```
andreamah, line 209: maybe add a comment that this "bytearray" isn't the same one as the constructor for the Image, but uses the system that requires you to convert to binary

andreamah (commented):
Awesome! Thanks for figuring this all out 😎

src/microbit/model/constants.py
vandyliu: will do

src/microbit/model/display.py
vandyliu: in python, if you have a 'try', you must have an 'except'

src/microbit/model/display.py
vandyliu: noted

src/microbit/model/display.py
vandyliu: removed the print now :P 
it's not really needed but yea you right

src/microbit/model/display.py
vandyliu: i updated it in the next commit, let me know what you think

src/microbit/model/display.py
vandyliu: Updated. Let me know what you think

andreamah (approved):
LGTM :D

src/microbit/model/display.py
```diff
@@ -0,0 +1,208 @@
...
+        while True:
+            if isinstance(value, (str, int, float)):
+                value = str(value)
+            else:
+                raise TypeError(f"can't convert {type(value)} object to str implicitly")
+            letters = []
+            for c in value:
+                if monospace:
+                    if c == " ":
+                        letters.append(Image(6, CONSTANTS.LED_HEIGHT))
```
nasadigital: Do we need to handle whitespace separately in the case of monospace?

src/microbit/model/display.py
```diff
@@ -0,0 +1,208 @@
...
+                        letters.append(Display.__get_image_from_char(c))
+                        letters.append(Image(1, CONSTANTS.LED_HEIGHT))
+                else:
+                    if c == " ":
+                        letters.append(Image(3, CONSTANTS.LED_HEIGHT))
+                    else:
+                        letters.append(
+                            Display.__strip_image(Display.__get_image_from_char(c))
+                        )
+                        letters.append(Image(1, CONSTANTS.LED_HEIGHT))
```
nasadigital: Can you put the constants in constant variables with descriptive names?

src/microbit/model/display.py
```diff
@@ -0,0 +1,208 @@
...
+        for row in image._Image__LED:
+            row.append(0)
+
+    @staticmethod
+    def __create_scroll_image(images):
+        blank_5x5_image = Image()
+        front_image = blank_5x5_image.crop(
+            0, 0, CONSTANTS.LED_WIDTH - 1, CONSTANTS.LED_HEIGHT
+        )
+        images.insert(0, front_image)
```
nasadigital: I think instantiating a 4x5 image using `Image(4, 5)` is much easier to understand compared to cropping a 5x5 image to 4x5.

src/microbit/model/display.py
```diff
@@ -0,0 +1,208 @@
...
+
+        scroll_image = Image._Image__append_images(images)
+        end_image = Image()
```
nasadigital: `front_image` and `end_image` sound very similar, making it seem like they are used in the same way (put `front_image` in the front, put `end_image` at the end) even though that is not the case.
Do you think you can make that distinction clearer?

src/microbit/model/display.py
```diff
@@ -0,0 +1,208 @@
...
+        # If c is not between the ASCII alphabet we cover, make it a question mark
+        if ord(c) < CONSTANTS.ASCII_START or ord(c) > CONSTANTS.ASCII_END:
+            c = "?"
+        offset = (ord(c) - CONSTANTS.ASCII_START) * CONSTANTS.LED_WIDTH
+        representative_bytes = CONSTANTS.ALPHABET[
+            offset : offset + CONSTANTS.LED_HEIGHT
+        ]
+        return Image(Display.__convert_bytearray_to_image_str(representative_bytes))
+
+    # Removes columns that are not lit
```
nasadigital, line 194: Maybe have the name of the function reflect this?
The current name sounds like it would strip both columns and rows.
if I see this function used in the wild, I would need to check the comment/docstring to find out that's not the case.

src/microbit/model/display.py
```diff
@@ -0,0 +1,208 @@
...
+        while True:
+            if isinstance(value, Image):
+                self.__image = value.crop(
+                    0, 0, CONSTANTS.LED_WIDTH, CONSTANTS.LED_HEIGHT
+                )
+            elif isinstance(value, (str, int, float)):
+                if isinstance(value, str):
+                    chars = list(value)
+                else:
+                    chars = list(str(value))
```
nasadigital: _&ast;nit&ast;_
I would drop the if clause and just have `chars = list(str(value))`. Up to you.

src/microbit/model/display.py
```diff
@@ -0,0 +1,208 @@
...
+
+                for c in chars:
+                    self.__image = Display.__get_image_from_char(c)
+                    time.sleep(delay / 1000)
+            else:
+                # Check if iterable
+                try:
+                    _ = iter(value)
+                except TypeError as e:
+                    raise e
```
nasadigital: I would probably do the prep for displaying outside of the loop and keep only the display stuff in the loop.
It's probably easier to understand and prevents logic duplication.

src/microbit/test/test_display.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+class TestDisplay(object):
+    def setup_method(self):
+        self.display = Display()
+
+    @pytest.mark.parametrize("x, y, brightness", [(1, 1, 4), (2, 3, 6), (4, 4, 9)])
+    def test_get_pixel(self, x, y, brightness):
+        self.display._Display__image._Image__LED[y][x] = brightness
+        assert brightness == self.display.get_pixel(x, y)
```
nasadigital: I would merge `test_get_pixel` and `test_set_pixel` into a single test and only test `display.set_pixel`/`display.get_pixel` instead of ` _Display__image._Image__LED`.
In general we want to focus on testing behavior, rather than implementation. If we change the implementation, ideally we don't want to have tests failing.

src/microbit/test/test_display.py
```diff
@@ -0,0 +1,148 @@
...
+        ],
+    )
+    def test_set_pixel_error(self, x, y, brightness, err_msg):
+        with pytest.raises(ValueError, match=err_msg):
+            self.display.set_pixel(x, y, brightness)
+
+    def test_clear(self):
+        self.display._Display__image._Image__LED[2][3] = 7
+        self.display._Display__image._Image__LED[3][4] = 6
+        self.display._Display__image._Image__LED[4][4] = 9
```
nasadigital: _&ast;nit&ast;_
I think I prefer `display.set_pixel` over `_Display__image._Image__LED`, since we don't expect users to directly access `__LED`. Up to you!

src/microbit/test/test_display.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def test_off(self):
+        self.display._Display__on = True
+        self.display.off()
+        assert False == self.display._Display__on
+
+    @pytest.mark.parametrize("on", [True, False])
+    def test_is_on(self, on):
+        self.display._Display__on = on
+        assert on == self.display.is_on()
```
nasadigital: Same as above. I would use `on()`, `off()` and `is_on()`, instead of changing the state directly.

src/microbit/model/display.py
```diff
@@ -0,0 +1,208 @@
...
+from .. import shim
+
+
+class Display:
+    # The implementation based off of https://github.com/bbcmicrobit/micropython/blob/master/docs/display.rst.
+    def __init__(self):
+        self.__image = Image()
+        self.__on = True
+
+    def scroll(self, value, delay=150, wait=True, loop=False, monospace=False):
```
nasadigital, line 21: Did we check if the behavior is the same on the device and here when you call `show/scroll(_, wait=False)` followed shortly by a `show/scroll(_, wait=True)`?

src/microbit/test/test_display.py
```diff
@@ -0,0 +1,148 @@
...
+        assert Image._Image__same_image(expected, self.display._Display__image)
+
+    def test_show_non_iterable(self):
+        with pytest.raises(TypeError):
+            self.display.show(TestDisplay())
+
+    def test_show_threaded(self):
+        threading.Thread = mock.Mock()
+        self.display.show("a", wait=False)
+        threading.Thread.assert_called_once()
```
nasadigital: I would remove this test (or even better change it so it tests behavior), since it tests the implementation.
For ex: if we switch to `async` in the future, this would fail even though maybe everything is working correctly.
Same for `test_scroll_threaded`.

nasadigital (commented):
Great job on making this faithful to the device! 👍

src/microbit/model/display.py
vandyliu: will remove unnecessary code :D

src/microbit/model/display.py
vandyliu: agreed

src/microbit/model/display.py
vandyliu: agree 100%

src/microbit/model/display.py
vandyliu: done !

src/microbit/model/display.py
vandyliu: changed the name

src/microbit/model/display.py
vandyliu: done

src/microbit/test/test_display.py
vandyliu: Agreed, changed

src/microbit/test/test_display.py
vandyliu: agreed

src/microbit/test/test_display.py
vandyliu: done

src/microbit/model/display.py
vandyliu: Changed in the next commit. Let me know what you think

src/microbit/test/test_display.py
vandyliu: I'm going to leave it for now but it is definitely something to look back to in the future.

src/microbit/model/display.py
vandyliu: Changed it to work the same as microbit device

src/microbit/model/display.py
```diff
@@ -58,38 +75,47 @@ def show(self, value, delay=400, wait=True, loop=False, clear=False):
...
-                        )
-                    elif isinstance(elem, str) and len(elem) == 1:
-                        self.__image = Display.__get_image_from_char(elem)
-                    # If elem is not char or image, break without iterating through rest of list
-                    else:
-                        break
+        while True:
+            for image in images:
+                self.__image = image
+                if use_delay:
```
nasadigital: I might just remove this variable and always use the delay, not sure it's worth the extra lines of code.
If the original library doesn't block when it's only one image to show, the check `if len(images) >= 2` could be used instead.

src/microbit/model/display.py
```diff
@@ -20,33 +23,47 @@ def scroll(self, value, delay=150, wait=True, loop=False, monospace=False):
...
             for x in range(appended_image.width() - CONSTANTS.LED_WIDTH + 1):
                 self.__image.blit(
                     appended_image, x, 0, CONSTANTS.LED_WIDTH, CONSTANTS.LED_HEIGHT
                 )
+                # If show or scroll is called again, there will be a different pid and break
+                if self.__current_pid != threading.get_ident():
+                    break
                 time.sleep(delay / 1000)
             if not loop:
                 break
```
nasadigital: I don't think the current implementation is thread-safe.
We should fix that :)

nasadigital (commented):
Thanks for applying the fixes! Just a bit more~

src/microbit/model/display.py
vandyliu: It blocks for 'delay' when it is an array of one image, but it doesnt block when it is just display.show(Image.HEART)

src/microbit/model/display.py
vandyliu: added locks 🥇

src/microbit/model/display.py
```diff
@@ -0,0 +1,256 @@
...
+                self.__image_lock.release()
+
+                # If show or scroll is called again, there will be a different pid and break
+                self.__thread_identifier_lock.acquire()
+                if self.__current_pid != threading.get_ident():
+                    break
+                self.__thread_identifier_lock.release()
+
+                if use_delay:
+                    time.sleep(delay / 1000)
```
xnkevinnguyen: nit: have a function to sleep so that you have the consistent /1000.

xnkevinnguyen: It's also going to make it clearer that all delays are passed as ms

src/microbit/model/display.py
```diff
@@ -58,13 +63,20 @@ def scroll(self, value, delay=150, wait=True, loop=False, monospace=False):
...
+                self.__image_lock.release()
+
                 # If show or scroll is called again, there will be a different pid and break
+                self.__thread_identifier_lock.acquire()
                 if self.__current_pid != threading.get_ident():
                     break
+                self.__thread_identifier_lock.release()
+
                 time.sleep(delay / 1000)
+
```
nasadigital: This method can still go wrong.
For e.g.: If we called `scroll(wait=False)` and after some time `show(wait=True)`, the following could happen:
`show` changes the image, `scroll` changes the image, `show` changes the pid, `scroll` changes the pid. On the screen it's a snapshot of scroll, instead of the image from show.
You could solve this by having one lock and doing the both pid and image checks/changes inside of it.

src/microbit/model/display.py
```diff
@@ -58,13 +63,20 @@ def scroll(self, value, delay=150, wait=True, loop=False, monospace=False):
...
                 self.__image.blit(
                     appended_image, x, 0, CONSTANTS.LED_WIDTH, CONSTANTS.LED_HEIGHT
                 )
+                self.__image_lock.release()
+
                 # If show or scroll is called again, there will be a different pid and break
+                self.__thread_identifier_lock.acquire()
                 if self.__current_pid != threading.get_ident():
                     break
+                self.__thread_identifier_lock.release()
```
nasadigital: If the condition is satisfied and we break, is `self.__thread_identifier_lock.release()` called?

src/microbit/model/display.py
```diff
@@ -13,8 +13,11 @@ class Display:
...
         self.__image = Image()
         self.__on = True
-        self.__blank_image = Image()
         self.__current_pid = None
+        self.__blank_image = Image()
+
+        self.__image_lock = threading.Lock()
+        self.__thread_identifier_lock = threading.Lock()
 
     def scroll(self, value, delay=150, wait=True, loop=False, monospace=False):
```
nasadigital, line 21: Do we know how `set_pixel` interacts with `scroll/show(wait=False)`?

src/microbit/model/microbit_model.py
```diff
@@ -9,6 +10,7 @@ def __init__(self):
         self.button_a = Button()
         self.button_b = Button()
         self.__start_time = time.time()
+        self.display = Display()
 
     def sleep(self, n):
         time.sleep(n / 1000)
```
xnkevinnguyen, line 16: Oh you have a sleep function here

src/microbit/model/display.py
vandyliu: Very true. Need to change that

src/microbit/model/display.py
vandyliu: will add lock to every part that accesses self.__image

nasadigital: Just curious, what would happen if you called `set_pixel` while it's scrolling?

vandyliu: Looks like it sets the pixel for a brief moment, but the scrolling keeps happening and the pixel is updated to whatever it should be while it is scrolling

src/microbit/model/microbit_model.py
vandyliu: dont want to use this one because of circular imports though, so i added a static function to display

src/microbit/model/display.py
vandyliu: will do

src/microbit/model/display.py
vandyliu: When you say 

> You could solve this by having one lock and doing the both pid and image checks/changes inside of it.

Do you mean have everything change in one block between acquire and release (Which is different from the change I just made)?

src/microbit/model/display.py
vandyliu: My implementation works the same way :D

src/microbit/model/display.py
```diff
@@ -63,19 +62,18 @@ def scroll(self, value, delay=150, wait=True, loop=False, monospace=False):
...
                 self.__image.blit(
                     appended_image, x, 0, CONSTANTS.LED_WIDTH, CONSTANTS.LED_HEIGHT
                 )
-                self.__image_lock.release()
 
                 # If show or scroll is called again, there will be a different pid and break
-                self.__thread_identifier_lock.acquire()
                 if self.__current_pid != threading.get_ident():
+                    self.__lock.release()
                     break
```
nasadigital: Perhaps consider doing this before updating the display? `__current_pid` might have been changed, but we check after the update.

nasadigital (approved):
Just address the last comment and LGTM! 🎉

src/microbit/model/display.py
nasadigital: Looks good to me in the last change.

src/microbit/model/display.py
vandyliu: Done 💃

PR #192, opened 2020-02-05
Led integration between webview and python api

xnkevinnguyen:
# Description:
Python code is compiled and can send messages to typescript api then to the webview.

- Led Listener for state changes on microbit
- Sending messages from the python side to the vscode api
- Send the valid messages from vscode api to the webview

## Type of change

# Limitations:

# Testing:

- [ ] Test A
- [ ] Test B

# Checklist:


src/common/utils.py
```diff
@@ -1,27 +1,28 @@
...
-from . import constants as CONSTANTS
-from . import debugger_communication_client
-from applicationinsights import TelemetryClient
+import sys
 
 previous_state = {}
 
 
-def show(state, debug_mode=False):
+def show(state, device_name, debug_mode=False):
```
nasadigital: I think this name is not very accurate. I think `emit`, `send_event` or something similar, would be more accurate.

src/microbit/model/constants.py
```diff
@@ -110,3 +117,5 @@
 NOT_IMPLEMENTED_ERROR = "This method is not implemented by the simulator"
 UNSUPPORTED_ADD_TYPE = "unsupported types for __add__:"
 SAME_SIZE_ERR = "images must be the same size"
+
+TIME_DELAY = 0.03
```
nasadigital: Maybe think about making this shorter? We shouldn't have that much logic and it would probably look smoother. Not urgent though.

src/view/components/microbit/MicrobitSimulator.tsx
```diff
@@ -0,0 +1,162 @@
...
+                    play_button: false,
+                });
+                break;
+            case "set-state":
+                console.log("Setting the state");
+                this.setState({
+                    leds: message.state.leds,
+                });
+                break;
+            case "activate-play":
```
nasadigital, line 55: I think maybe something like `confirm-start` or `confirm-play` will make this event more clear to the reader.

src/view/components/microbit/MicrobitSimulator.tsx
```diff
@@ -0,0 +1,162 @@
...
+        );
+    }
+    protected togglePlayClick = () => {
+        sendMessage("play-simulator", {
+            active_device: CONSTANTS.DEVICE_NAME.MICROBIT,
+            selected_file: this.state.selected_file,
+            state: !this.state.play_button,
+        });
+    };
+    protected onSelectBlur(event: React.FocusEvent<HTMLSelectElement>) {
```
nasadigital: How is this different from `onSelect`? If you need a unique function name maybe something like `onSelectFile` could work?

src/extension.ts
```diff
@@ -530,6 +563,7 @@ export async function activate(context: vscode.ExtensionContext) {
                 if (currentPanel) {
                     console.log("Sending clearing state command");
                     currentPanel.webview.postMessage({
+                        active_device: currentActiveDevice,
                         command: "reset-state",
```
vandyliu, line 567: this is a comment for the future but should we have these commands as constants somewhere?

src/extension.ts
```diff
@@ -1050,6 +1088,10 @@ function getWebviewContent(context: vscode.ExtensionContext) {
           </body>
           </html>`;
 }
+function switchDevice(deviceName: string) {
+    console.log("Switched Devices");
```
vandyliu: should we keep this console.log?

src/common/utils.py
```diff
@@ -1,27 +1,28 @@
...
-    if state != previous_state:
-        previous_state = copy.deepcopy(state)
+
+    if not device_name in previous_state or state != previous_state[device_name]:
+
+        previous_state[device_name] = copy.deepcopy(state)
+        state_ext = {
+            "device_name": device_name,
+        }
+        state.update(state_ext)
```
nasadigital: Can we put this outside of the if? If we can, I think it would streamline this method a bit.

nasadigital (commented):
Great progress!!

src/view/components/microbit/MicrobitSimulator.tsx
```diff
@@ -0,0 +1,162 @@
...
+                this.setState({
+                    ...this.state,
+                    leds: initialLedState,
+                    play_button: false,
+                });
+                break;
+            case "set-state":
+                console.log("Setting the state");
+                this.setState({
+                    leds: message.state.leds,
```
vandyliu, line 52: this one is missing ...this.state

Also, are they even necessary? I thought setState patches the fields that are given to it.

src/process_user_code.py
```diff
@@ -43,7 +44,7 @@ def run(self):
             sys.stdin.flush()
             try:
                 new_state = json.loads(read_val)
-                for event in CONSTANTS.EXPECTED_INPUT_EVENTS:
+                for event in CONSTANTS.EXPECTED_INPUT_EVENTS_CPX:
```
vandyliu, line 47: Is process_user_code going to be used for both cpx and microbit?

src/common/utils.py
andreamah: I put it all in a create_message helper function after fixing some stuff up (about to commit it). Does that help out? Do you mean streamline as in with clarity or runtime?

src/process_user_code.py
andreamah: yes, it's not in either library, so I'm assuming that it will be common

src/microbit/model/constants.py
andreamah: Perhaps a bit later when we have cleaned everything up 😊. Thanks for the suggestion!!!

src/common/utils.py
andreamah: ah ok thanks :D

src/view/components/microbit/MicrobitSimulator.tsx
xnkevinnguyen: Yea, they don't seem to be necessary. I usually don't use it unless i'm modifying an object inside the state and I want to keep the attributes the same.

src/common/utils.py
```diff
@@ -1,27 +1,28 @@
...
         message = {"type": "state", "data": json.dumps(state)}
-        if debug_mode:
```
andreamah: note: currently trying to restore debug for cpx... accidentally removed it

src/common/utils.py
andreamah: I replied to this in the file changes area, but it doesn't seem to be here... :( here's what I said:
I put it all in a create_message helper function after fixing some stuff up (about to commit it). Does that help out? Do you mean streamline as in with clarity or runtime?

nasadigital: Sorry, the comment was unclear! Chatted offline to elaborate what I meant.

package.json
```diff
@@ -1,365 +1,366 @@
...
-    "vscode-extension-telemetry": "^0.1.1",
-    "vscode-nls": "^4.1.0"
-  },
-  "eslintConfig": {
-    "extends": "react-app"
-  },
-  "extensionDependencies": [
-    "ms-python.python"
-  ]
 }
```
sagarmanchanda, line 366: Any reason for this re ordering?

src/view/App.tsx
```diff
@@ -9,6 +9,14 @@ import { Tab } from "./components/tab/Tab";
 import { DEVICE_LIST_KEY } from "./constants";
 import { Device } from "./container/device/Device";
 
+interface vscode {
+    postMessage(message: any): void;
+}
+declare const vscode: vscode;
+
+const sendMessage = (type: string, state: any) => {
```
sagarmanchanda: I'd abstract this out as util somewhere and use it all the places

src/view/components/cpx/CpxSimulator.tsx
```diff
@@ -65,9 +62,9 @@ const sendMessage = (type: string, state: any) => {
     vscode.postMessage({ command: type, text: state });
 };
 
-class Simulator extends React.Component<any, IState> {
-    constructor(props: IMyProps) {
-        super(props);
+class Simulator extends React.Component<{}, IState> {
+    constructor() {
```
sagarmanchanda: constructor(props) {
    super(props);
}

sagarmanchanda: Don't need the type here, It'll automatically infer from your React.Component definition

src/view/components/cpx/CpxSimulator.tsx
```diff
@@ -88,6 +85,9 @@ class Simulator extends React.Component<any, IState> {
 
     handleMessage = (event: any): void => {
         const message = event.data; // The JSON data our extension sent
+        if (message.active_device !== DEVICE_LIST_KEY.CPX) {
+            return;
+        }
```
sagarmanchanda, line 82: Does the python communication bridge run all components handleMessage? If yes, it is worth changing at the communication level rather than checking here.

src/view/components/microbit/MicrobitImage.tsx
```diff
@@ -5,6 +5,81 @@ import * as React from "react";
...
+        button: HTMLElement,
+        event: Event,
+        buttonKey: string
+    ) => void;
+}
+interface IProps {
+    eventTriggers: EventTriggers;
+    leds: number[][];
+}
+interface IState {
```
sagarmanchanda: Refs should not be a part of the state, I'm assuming the ref would stay constant throughout the lifecycle of the component

src/view/components/microbit/MicrobitSimulator.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+interface vscode {
+    postMessage(message: any): void;
+}
+
+declare const vscode: vscode;
+
+const sendMessage = (type: string, state: any) => {
+    vscode.postMessage({ command: type, text: state });
+};
```
sagarmanchanda: duplication

src/view/components/microbit/MicrobitSimulator.tsx
```diff
@@ -0,0 +1,157 @@
...
+    }
+    handleMessage = (event: any): void => {
+        const message = event.data;
+        if (message.active_device !== DEVICE_LIST_KEY.MICROBIT) {
+            return;
+        }
+
+        switch (message.command) {
+            case "reset-state":
+                console.log("Reset the state");
```
sagarmanchanda: nit: Console.logs present throughout the PR

src/view/components/microbit/MicrobitSimulator.tsx
```diff
@@ -0,0 +1,157 @@
...
+    postMessage(message: any): void;
+}
+
+declare const vscode: vscode;
+
+const sendMessage = (type: string, state: any) => {
+    vscode.postMessage({ command: type, text: state });
+};
+interface IState {
+    active_editors: string[];
```
sagarmanchanda, line 19: Lets review everything that is a part of state in person, We should be very careful while deciding what to put in state as a change of same leads to re render and we don't want unnecessary renders

sagarmanchanda: Also sometimes it is good to define the shouldcomponentupdate in your component

sagarmanchanda (changes requested):

src/view/components/cpx/CpxSimulator.tsx
xnkevinnguyen: It doesn't infer it automatically since I have no implicit any rule. I can do this instead:
    constructor(props: Readonly<{}>) {
        super(props);

src/view/App.tsx
xnkevinnguyen: Will do

src/view/components/cpx/CpxSimulator.tsx
xnkevinnguyen: Will do. Will verify before sending set-state from vscode api

src/extension.ts
xnkevinnguyen: Yea, it makes sense. I will refactor all those messages in a separate pr.

src/view/components/microbit/MicrobitSimulator.tsx
xnkevinnguyen: will be in a util

src/view/components/microbit/MicrobitSimulator.tsx
xnkevinnguyen: I will put all those messages in a constant, a name I would put is toggle-play-stop

package.json
andreamah: I accidentally re-ordered when resolving merge conflicts :( is it ok if we just leave it like this?

src/adafruit_circuitplayground/constants.py
```diff
@@ -7,6 +7,8 @@
 
 BRIGHTNESS_RANGE_ERROR = "The brightness value should be a number between 0 and 1."
 
+CPX = "CPX"
```
sagarmanchanda, line 10: Putting it inside a device?

src/view/components/cpx/CpxSimulator.tsx
```diff
@@ -119,14 +106,12 @@ class Simulator extends React.Component<any, IState> {
...
                 );
                 this.setState({
-                    ...this.state,
                     active_editors: message.state.activePythonEditors,
                 });
                 break;
             case "current-file":
                 console.log("Setting current file", message.state.running_file);
                 this.setState({
-                    ...this.state,
```
sagarmanchanda, line 129: +1 for cleanup :D

src/view/components/microbit/MicrobitImage.tsx
```diff
@@ -3,8 +3,83 @@
...
+    microbitImageRef: React.RefObject<SVGSVGElement>;
```
sagarmanchanda: Why are these still in the state?

src/view/components/microbit/MicrobitImage.tsx
```diff
@@ -3,8 +3,83 @@
...
+}
+
+// Displays the SVG and call necessary svg modification.
+export class MicrobitImage extends React.Component<IProps, IState> {
+    private svgRef: React.RefObject<MicrobitSvg> = React.createRef();
+    constructor(props: IProps) {
+        super(props);
+    }
+    componentDidMount() {
+        const svgElement = this.svgRef.current;
```
sagarmanchanda, line 29: null check missing this.svgRef && this.svgRef.current

src/view/components/microbit/MicrobitImage.tsx
```diff
@@ -3,8 +3,83 @@
...
+        if (svgElement) {
+            updateAllLeds(this.props.leds, svgElement.getLeds());
+            setupAllButtons(this.props.eventTriggers, svgElement.getButtons());
+        }
+    }
+    componentDidUpdate() {
+        if (this.svgRef.current) {
```
sagarmanchanda, line 36: null check missing this.svgRef

src/view/components/microbit/MicrobitImage.tsx
```diff
@@ -3,8 +3,83 @@
...
+        }
+    }
+};
+const updateAllLeds = (
+    leds: number[][],
+    ledRefs: Array<React.RefObject<SVGRectElement>>[]
+) => {
+    for (let j = 0; j < leds.length; j++) {
+        for (let i = 0; i < leds[0].length; i++) {
+            const ledElement = ledRefs[j][i].current;
```
sagarmanchanda, line 72: I would do a null check for ledRefs here just to be sure

src/view/components/microbit/MicrobitImage.tsx
```diff
@@ -3,8 +3,83 @@
...
+            if (ledElement) {
+                setupLed(ledElement, leds[i][j]);
+            }
+        }
+    }
+};
+const setupLed = (ledElement: SVGRectElement, brightness: number) => {
```
sagarmanchanda, line 79: nit: typo SVGReactElement, can you update all the instances

src/view/components/microbit/MicrobitSimulator.tsx
```diff
@@ -0,0 +1,138 @@
...
+                    <MicrobitImage
+                        eventTriggers={{
+                            onMouseDown: this.onMouseDown,
+                            onMouseUp: this.onMouseUp,
+                            onMouseLeave: this.onMouseLeave,
+                        }}
+                        leds={this.state.leds}
+                    />
+                </div>
+                <ActionBar
```
sagarmanchanda, line 107: This Action bar is separate from the one at the root level right?

src/view/utils/MessageUtils.tsx
```diff
@@ -0,0 +1,9 @@
+interface vscode {
+    postMessage(message: any): void;
```
sagarmanchanda, line 2: Is it worth typifying this?

src/view/utils/MessageUtils.tsx
```diff
@@ -0,0 +1,9 @@
+interface vscode {
+    postMessage(message: any): void;
+}
+
+declare const vscode: vscode;
+
+export const sendMessage = (type: string, state: any) => {
```
sagarmanchanda: I'd use generics here:

export const sendMessage\<TState\> = (type: string, state: TState) => { ... }

sagarmanchanda (changes requested):

package.json
sagarmanchanda: yeah sure no issues

src/adafruit_circuitplayground/constants.py
vandyliu: like an enum?

src/view/components/microbit/MicrobitSimulator.tsx
xnkevinnguyen: the root level is the tab, the action bar is toggle play

src/adafruit_circuitplayground/constants.py
andreamah: Based on my research, it seems that you usually don't put the constants in objects in python (https://stackoverflow.com/questions/6345840/whats-the-best-way-to-initialise-and-use-constants-across-python-classes). This was also how the last cohort arranged their constants.

sagarmanchanda: fell free to resolve this

src/view/components/microbit/MicrobitSimulator.tsx
sagarmanchanda: sounds good

src/view/components/cpx/CpxSimulator.tsx
sagarmanchanda: sounds good. feel free to resolve this

src/view/components/microbit/MicrobitImage.tsx
xnkevinnguyen: Updated the typing

src/view/components/microbit/MicrobitImage.tsx
xnkevinnguyen: I've typed svgRef as always defined since I create it in the same class before the component mounts

src/view/components/microbit/MicrobitImage.tsx
xnkevinnguyen: same here, svgRef should always be defined

src/view/components/microbit/MicrobitImage.tsx
xnkevinnguyen: SVGRectElement are referencing to tsx svg rect elements. For example,   
 <rect
                            className="sim-button-outer"
                            ref={this.buttonRefs.BTN_B}
                            x="418.1"
                            y="176.4"
                            rx={4}
                            ry={4}
                            width="56.2"
                            height="56.2"
                        />

src/view/utils/MessageUtils.tsx
xnkevinnguyen: Good tip, changed to generics

src/view/utils/MessageUtils.tsx
xnkevinnguyen: this interface links to the vscode api which has any type. The only place where we call this specific interface is in sendMessage (with generic typing in the following comment you made)

src/common/utils.py
```diff
@@ -10,17 +10,23 @@
 previous_state = {}
 
 
-def show(state, device_name, debug_mode=False):
+def create_message(state, device_name):
+    state_ext = {
+        "device_name": device_name,
+    }
+    state.update(state_ext)
```
nasadigital: I don't think we should change the input parameter `state` inside this method. Making a copy, updating the copy and returning it is probably the way to go. This behavior can be separated from the `create_message` function (wouldn't expect a method with that name to also have the side effect of changing the input parameter).
Sorry for the multiple change requests on this part!

nasadigital (approved):
LGTM! Great job everyone 🥇

src/common/utils.py
```diff
@@ -1,27 +1,35 @@
...
+    message = {"type": "state", "data": json.dumps(state_copy)}
+    return message
+
+
+def send_to_simulator(state, device_name):
     global previous_state
+
+    message = create_message(state, device_name)
+
     if state != previous_state:
```
nasadigital: Thanks for addressing the comment!
But now this part might not work as expected, right?
I would take the state updating logic and put it in this method (or in a function and call that function here).

sagarmanchanda (approved):

sagarmanchanda:
Looks good. Amazing work. 💯

src/common/utils.py
andreamah: Thanks! I just abstracted out the state updating and made the process two function calls :).

nasadigital: Awesome!
One more thing, now `send_to_simulator` and `debug_show` mutate the input parameters. It would be easier for a developer if they can use these functions without worrying about any unintended consequences.

andreamah: ok, tried to fix it in the last commit :)

nasadigital: Almost perfect!
Now this check will always return true `if state != previous_state:`. Perhaps use `state_copy` instead of `state`?

andreamah: oops! fixed

nasadigital: _&ast;nit&ast;_
I would change the name to `updated_state` from `state_copy`. Up to you!

andreamah: fixed!

nasadigital: LGTM!
Thanks for your patience :D
giphy

PR #195, opened 2020-02-10
Button interaction

xnkevinnguyen:
# Description:
Microbit button should be fully integrated with the python code. 

## Type of change


src/process_user_code.py
```diff
@@ -43,16 +46,39 @@ def run(self):
...
-                        event, cpx._Express__state[event]
-                    )
+                new_state_message = json.loads(read_val)
+                device = new_state_message.get(CONSTANTS.ACTIVE_DEVICE_FIELD)
+                new_state = new_state_message.get(CONSTANTS.STATE_FIELD, {})
+
+                if device == CPX:
+                    update_cpx(new_state)
+                elif device == MICROBIT:
+                    update_microbit(new_state)
```
isadorasophia: Is it possible to get a parent "UserInput" class and inherit classes specific to CPX and microbit? This would make things easier to scale in the future. We wouldn't need to check for the active device in every iteration, etc.

isadorasophia (changes requested):
yay! buttons :D

sagarmanchanda (approved):
Looks good. :)

src/process_user_code.py
andreamah: It gets a little tricky to instantiate a specific child class based on the device, since the device that the tab is currently on is not known until the first interaction (in this case, the button press) happens. However, the UserInput class is created when the script first starts. If we go for something like that, we would need to pass in an extra argument when we start the python backend based on which tab they are on when they press play, which may take a little time. 
Therefore, I thought that this if..else... was okay given the current structure. What do you think? Sorry for the long read!

andreamah: Technically, we can see which libraries import, but in theory, both libraries could be imported. In terms of keeping track of which device is in use, we're using whichever tab the frontend is on, which is only really kept track of on the frontend.

src/process_user_code.py
```diff
@@ -43,16 +46,39 @@ def run(self):
...
+    for button in CONSTANTS.EXPECTED_INPUT_BUTTONS_MICROBIT:
+        previous_pressed = None
+        exec(f"previous_pressed = mb.{button}.get_presses()")
+        button_pressed = new_state.get(button, previous_pressed)
+
+        if button_pressed != previous_pressed:
+            if button_pressed:
+                exec(f"mb.{button}._Button__press_down()")
+            else:
+                exec(f"mb.{button}._Button__release()")
```
nasadigital: Can we use a dict to get the correct button object instead of using `exec`?

src/process_user_code.py
```diff
@@ -43,16 +46,39 @@ def run(self):
...
 
+def update_cpx(new_state):
+    for event in CONSTANTS.EXPECTED_INPUT_EVENTS_CPX:
+        cpx._Express__state[event] = new_state.get(event, cpx._Express__state[event])
+
+
+def update_microbit(new_state):
+    for button in CONSTANTS.EXPECTED_INPUT_BUTTONS_MICROBIT:
+        previous_pressed = None
+        exec(f"previous_pressed = mb.{button}.get_presses()")
```
nasadigital: I think you meant to use `is_pressed()` here?
`get_presses()` resets the "button clicked" counter and we probably want to refrain from using it in the backend.
From the method docstring:
```
Returns the running total of button presses, and resets this total
to zero before returning.
```

src/process_user_code.py
isadorasophia: Got it! Do we have some kind of event whenever the button for switching gets pressed? It would be nice if we could at least save some kind of state at UserInput. 

If that ends up being too much work we can prioritize that for later too, I was mostly worried about an overhead/further scaling.

nasadigital (commented):
Very neat PR. Good job!

src/process_user_code.py
```diff
@@ -43,16 +46,39 @@ def run(self):
...
             except Exception as e:
                 print(CONSTANTS.ERROR_SENDING_EVENT, e, file=sys.stderr, flush=True)
 
 
+def update_cpx(new_state):
+    for event in CONSTANTS.EXPECTED_INPUT_EVENTS_CPX:
+        cpx._Express__state[event] = new_state.get(event, cpx._Express__state[event])
+
+
+def update_microbit(new_state):
```
nasadigital: I think it would make more sense to handle the state changing logic inside each component, instead of having all the different components update state in this method. What do you think?

src/process_user_code.py
andreamah: oops, called the wrong function! thanks for catching that!

src/process_user_code.py
andreamah: that's true... I'll try implementing that for the next commit!

src/process_user_code.py
andreamah: We have it so that, if the user presses another tab mid-run, the process stops. This is controlled via the frontend.

Also, I just made a commit that sorta abstracts the device updating away (for better scaling). Now, the updating method is within the microbit/cpx class itself, as per David's comment below! :)

src/process_user_code.py
andreamah: implemented!

src/process_user_code.py
xnkevinnguyen: When we switch tab, we kill the process. We also tag each message with active_device, so that we can filter out the messages if the device doesn't match.

isadorasophia: sounds good!

isadorasophia (approved):

PR #196, opened 2020-02-10
Added light, temperature and accelerometer on python side

vandyliu:
# Description:

Added light, temperature and accelerometer on python side

## Type of change

# Limitations:

# Testing:

- [ ] Test A
- [ ] Test B

# Checklist:


src/microbit/__model/display.py
```diff
@@ -226,13 +227,19 @@ def is_on(self):
...
-
         Use the display's LEDs in reverse-bias mode to sense the amount of light
         falling on the display.  Returns an integer between 0 and 255 representing
         the light level, with larger meaning more light.
         """
-        raise NotImplementedError(CONSTANTS.NOT_IMPLEMENTED_ERROR)
+        return self.__light_level
+
+    def __set_light_level(self, level):
+        if level < CONSTANTS.MIN_LIGHT_LEVEL:
```
xnkevinnguyen: Is this how we should handle out of range values? Or should we throw an error

vandyliu: Great question. Not sure to be honest. On the CPX, if I inputted a value less than the min (on the simulator), it defaulted to the min value, which is why I had it here.

xnkevinnguyen: I see, maybe one of the coaches has advice on this. I am not sure either. I feel like it could "hide" some incorrect inputs if that ever happens

src/microbit/test/test_shim.py
```diff
@@ -20,3 +20,8 @@ def test_running_time(self):
         MicrobitModel.running_time = mock.Mock()
         running_time()
         MicrobitModel.running_time.assert_called_once()
+
```
andreamah: Perhaps we should use this PR to consider renaming test_shim.py? Since there's no more shim, the name doesn't really make sense.

src/microbit/__model/microbit_model.py
```diff
@@ -1,23 +1,39 @@
...
         time.sleep(n / 1000)
 
     def running_time(self):
         print(f"time. time: {time.time()}")
         return time.time() - self.__start_time
 
+    def temperature(self):
+        return self.__temperature
+
+    def __set_temperature(self, temperature):
```
andreamah, line 35: Considering that you restrain the result here and with light level, perhaps it would be helpful to implement some sort of helper function to restrain it? You can probably put it in common.

src/microbit/__model/accelerometer.py
```diff
@@ -0,0 +1,129 @@
+import enum
```
andreamah: why do you still import enum here?

andreamah (commented):
Yay! Thanks for all this work :D

src/microbit/__model/accelerometer.py
vandyliu: awks, it shall disappear

src/microbit/test/test_shim.py
vandyliu: changed to test_init.py

src/microbit/__model/microbit_model.py
andreamah: https://stackoverflow.com/questions/5996881/how-to-limit-a-number-to-be-within-a-specified-range-python

src/microbit/__model/display.py
isadorasophia: I think that we want to do make sense of whatever the real thing does. It wouldn't read a value that is below it's minimum level -- so I'd suggest throwing an exception if that's the case.

vandyliu: Okay, I will update it to raise an exception :)

src/microbit/__model/microbit_model.py
vandyliu: I would have done that but the implementation has changed so I won't be using this ):
I'm definitely saving this though! Thank you 💃

andreamah (approved):
Looks good to me :D

src/microbit/__model/accelerometer.py
```diff
@@ -0,0 +1,112 @@
...
+    def was_gesture(self, name):
+        """
+        Return True or False to indicate if the named gesture was active since the
+        last call.
+        """
+        self.__add_current_gesture_to_gesture_lists()
+        if name not in CONSTANTS.GESTURES:
+            raise ValueError(CONSTANTS.INVALID_GESTURE_ERR)
+        was_gesture = name in self.__prev_gestures
+        self.__prev_gestures.clear()
```
isadorasophia, line 70: Is there a reason we are clearing this? The API doesn't seem to suggest that we lose the previous state

src/microbit/__model/accelerometer.py
```diff
@@ -0,0 +1,112 @@
...
+        elif axis == "y":
+            self.__y = accel
+        elif axis == "z":
+            self.__z = accel
+
+    def __set_gesture(self, gesture):
+        if gesture in CONSTANTS.GESTURES:
+            self.__current_gesture = gesture
+        else:
+            self.__current_gesture = ""
```
isadorasophia, line 107: Raise an error?

isadorasophia (approved):
🏄

src/microbit/__model/accelerometer.py
vandyliu: prev_gestures is to remember all gestures that have existed since the last was_gesture call.
From the documentation, once the function is called, all the gestures in prev_gestures disappear (if that makes sense).

src/microbit/__model/accelerometer.py
vandyliu: will do

src/microbit/test/test_microbit_model.py
```diff
@@ -20,7 +21,21 @@ def test_running_time(self):
...
+    @pytest.mark.parametrize(
+        "temperature", [CONSTANTS.MIN_TEMPERATURE, 0, CONSTANTS.MAX_TEMPERATURE],
+    )
+    def test_temperature(self, temperature):
+        self.__mb._MicrobitModel__set_temperature(temperature)
+        assert temperature == self.__mb.temperature()
+
+    @pytest.mark.parametrize(
+        "invalid_temperature",
+        [CONSTANTS.MIN_TEMPERATURE - 10, CONSTANTS.MAX_TEMPERATURE + 5],
```
nasadigital: _&ast;nit&ast;_
I would do `CONSTANTS.MIN_TEMPERATURE - 1` and `CONSTANTS.MAX_TEMPERATURE + 1` to check the corner cases instead.

nasadigital (approved):
LGTM ⚡️⚡️⚡️

src/microbit/test/test_microbit_model.py
vandyliu: will change then merge! Thanks David

PR #216, opened 2020-02-20
Basic functionality for microbit Debugging and fixes

andreamah:
# Description:

Improvements to CPX debugger and addition of microbit debugger.

## Type of change

# Limitations:

# Testing:

# Checklist:


src/common/debugger_communication_client.py
```diff
@@ -0,0 +1,81 @@
...
+
+    except Exception as e:
+        print(CONSTANTS.ERROR_SENDING_EVENT, e, file=sys.stderr, flush=True)
+
+
+# Method : Update State
+def update_state(state):
+    processing_state_event.clear()
+    sio.emit("updateState", state)
+    processing_state_event.wait()
```
isadorasophia, line 68: Do you know what happens if we terminate the debugging session before this gets set? (I'm assuming the process gets terminated, just want to confirm that!)

src/debuggerCommunicationServer.ts
```diff
@@ -36,16 +55,22 @@ export class DebuggerCommunicationServer {
...
-        console.log(`Emit Sensor Changed: ${newState} \n`);
-        this.serverIo.emit("sensor_changed", newState);
+    public emitInputChanged(newState: string): void {
+        if (this.isWaitingResponse) {
+            this.currentCall.push(() => {
+                this.serverIo.emit(
+                    DEBUGGER_MESSAGES.EMITTER.INPUT_CHANGED,
+                    newState
+                );
+                this.isWaitingResponse = true;
```
isadorasophia: Do you know if this runs single-threaded? (I _think_ it does?) If not, we might want to consider atomic operations when getting/setting the value of `isWaitingResponse`.

src/debuggerCommunicationServer.ts
```diff
@@ -6,15 +6,32 @@ import * as socketio from "socket.io";
...
+};
+
 export class DebuggerCommunicationServer {
     private port: number;
     private serverHttp: http.Server;
     private serverIo: socketio.Server;
     private simulatorWebview: WebviewPanel | undefined;
+    private currentActiveDevice;
+    private isWaitingResponse = false;
+    private currentCall: Array<Function> = [];
```
isadorasophia: nit: maybe we can rename this for `pendingCallbacks`? Or something in these lines

src/debuggerCommunicationServer.ts
```diff
@@ -36,16 +55,22 @@ export class DebuggerCommunicationServer {
...
-    }
-
-    // Emit Sensors Inputs Events
-    public emitSensorChanged(newState: string): void {
-        console.log(`Emit Sensor Changed: ${newState} \n`);
-        this.serverIo.emit("sensor_changed", newState);
+    public emitInputChanged(newState: string): void {
+        if (this.isWaitingResponse) {
+            this.currentCall.push(() => {
+                this.serverIo.emit(
```
isadorasophia, line 61: I would add a comment here explaining that since we already have an active event, we will just push this to our queue until that event is processed.

src/debuggerCommunicationServer.ts
```diff
@@ -6,15 +6,32 @@ import * as socketio from "socket.io";
...
+    },
+};
+
 export class DebuggerCommunicationServer {
     private port: number;
     private serverHttp: http.Server;
     private serverIo: socketio.Server;
     private simulatorWebview: WebviewPanel | undefined;
+    private currentActiveDevice;
+    private isWaitingResponse = false;
```
isadorasophia: nit: similar to the comment below, we can change this to `isPendingResponse`? It's up to you though

src/debuggerCommunicationServer.ts
```diff
@@ -57,14 +82,23 @@ export class DebuggerCommunicationServer {
...
-            socket.on("updateState", (data: any) => {
+            socket.on(DEBUGGER_MESSAGES.LISTENER.UPDATE_STATE, (data: any) => {
                 this.handleState(data);
+                this.serverIo.emit(
+                    DEBUGGER_MESSAGES.EMITTER.RECEIVED_STATE,
+                    {}
+                );
+            });
+            socket.on(DEBUGGER_MESSAGES.LISTENER.RECEIVED_STATE, () => {
+                this.isWaitingResponse = false;
```
isadorasophia: We can alternatively only assign this to false if `this.currentCall.length != 0` -- then we wouldn't need to update this value inside `currentCall()` again.

isadorasophia (changes requested):
Nice job everyone! 🐊

src/common/debugger_communication_client.py
andreamah: Yup, I tried stopping the session right in the middle and it still works :). There may be some issues with restarting without stopping first, but it is known and will be fixed in future commits in #217

src/debuggerCommunicationServer.ts
xnkevinnguyen: Good suggestions! it is much better naming

src/debuggerCommunicationServer.ts
xnkevinnguyen: Yes, since we're not using async this is single threaded

src/debuggerCommunicationServer.ts
xnkevinnguyen: good suggestion! Removed it from current call and setting it to true if this.currentcall.lenth>0. If it's 0, then i'm putting it to false since there's no more calls

src/common/debugger_communication_client.py
isadorasophia: Cool!

src/debuggerCommunicationServer.ts
isadorasophia: I _think_ we don't need to assign it to true on the listener? Since we won't assign it to false elsewhere. Anyway, it's up to you if you think that's a good sanity check!

isadorasophia (approved):
Just a heads up, try to keep only one change per PR (e.g. I noticed we have some stuff for telemetry as well). It's easier to roll back if there is any breaking changes if we keep one thing per insertion.
I think that's all right for now, just to keep that in mind for the next ones!

PR #218, opened 2020-02-22
Virtual Environment Support for Dependencies

andreamah:
# Description:

Uses a virtual environment in extension files for holding adafruit_circuitplayground dependencies. Also allows for users to use their own virtual environments + install dependencies to those ones. 

I also added a new option in settings to automatically configure environments when you switch the python interpreter. It is set to false by default, as it would probably be pretty disruptive if you're not using the extension. 

There still the command palette option to configure the environment, which is now adapted for the new workflow.

If the python interpreter option ever gets deprecated and the "python.pythonPath" setting is no longer available, it will just use the global python version to create a virtual environment in the extension files every time.

Former dependencyChecker.ts system is deleted, since we are now more reliant on the python interpreter's ability to find the python path rather than what the python path environment variable is pointing to.

If virtual environment creation on the extension side ever fails, the extension asks the user if it can just install dependencies to whichever python interpreter they have selected & it stays on that interpreter.

If the user already has all dependencies installed on the python version that the interpreter is pointing to, it doesn't care whether they're in a venv and just goes with the current interpreter with no setup.

## Type of change

# Limitations:

# Testing:

# Checklist:


.gitignore
```diff
@@ -8,6 +8,9 @@ out/
 !locales/**/out/  
 package.nls.*.json
 
+# virtual environment
+env/
```
vandyliu: What are your thoughts on `venv/` vs `env/`

andreamah: It doesn't really matter! If you prefer venv/ as the folder, I can change it.

andreamah: changed!

src/check_if_venv.py
```diff
@@ -0,0 +1,8 @@
+# from https://stackoverflow.com/questions/1871549/determine-if-python-is-running-inside-virtualenv
+import sys
+
+isVenv = hasattr(sys, "real_prefix") or (
+    hasattr(sys, "base_prefix") and sys.base_prefix != sys.prefix
+)
+
+print(int(isVenv))
```
xnkevinnguyen, line 11: Is this something you want to print?

src/constants.ts
```diff
@@ -79,6 +79,10 @@ export const CONSTANTS = {
             "error.invalidFileExtensionDebug",
             "The file you tried to run isn't a Python file."
         ),
+        INVALID_PYTHON_PATH: localize(
+            "error.invalidPythonPath",
+            'We found that your selected Python interpreter version is too low to run the extension. Please upgrade to version 3.7+ or select a different interpreter (CTRL+SHIFT+p and type "python.selectInterpreter") and restart the application.'
```
xnkevinnguyen: nit: CTRL + SHIFT + P instead of CTRL + SHIFT + p

locales/en/package.i18n.json
```diff
@@ -12,5 +12,7 @@
   "deviceSimulatorExpressExtension.commands.microbit.openSimulator": "[micro:bit] Open Simulator",
   "deviceSimulatorExpressExtension.commands.microbit.newFile": "[micro:bit] New File",
   "deviceSimulatorExpressExtension.configuration.title": "Device Simulator Express configuration",
-  "deviceSimulatorExpressExtension.configuration.properties.debuggerPort": "The port the Server will listen on for communication with the debugger."
-}
+  "deviceSimulatorExpressExtension.configuration.properties.configEnvOnChange": "When you change the Python Interpreter, the dependencies for the Device Simulator Express will automatically be configured for the extension dependencies (and the interpreter will auto-change if necessary).",
```
xnkevinnguyen: It might be confusing for the users if you add that the interpreter will auto change if necessary, but then later you ask them to ctrl shift p to change the interpreter manually

xnkevinnguyen (changes requested):

locales/en/package.i18n.json
nasadigital: This message is not very clear, I read it multiple times and I still don't quite understand what it does.

package.nls.json
```diff
@@ -12,5 +12,7 @@
   "deviceSimulatorExpressExtension.commands.microbit.openSimulator": "[micro:bit] Open Simulator",
   "deviceSimulatorExpressExtension.commands.microbit.newFile": "[micro:bit] New File",
   "deviceSimulatorExpressExtension.configuration.title": "Device Simulator Express configuration",
-  "deviceSimulatorExpressExtension.configuration.properties.debuggerPort": "The port the Server will listen on for communication with the debugger."
-}
+  "deviceSimulatorExpressExtension.configuration.properties.configEnvOnChange": "When you change the Python Interpreter, the dependencies for the Device Simulator Express will automatically be configured for the extension dependencies (and the interpreter will auto-change if necessary).",
+  "deviceSimulatorExpressExtension.configuration.properties.debuggerPort": "The port the Server will listen on for communication with the debugger.",
+  "deviceSimulatorExpressExtension.configuration.properties.dependencyChecker": "Whether or not to ask for dependency downloads. If unchecked, the extension will default to never download dependencies, except when automatically creating a virtual environment in the extension files."
```
nasadigital: I don't think this description is shown anywhere.

src/check_if_venv.py
andreamah: Yes, because the frontend would read the result this way.

locales/en/package.i18n.json
andreamah: Everything only works if the original interpreter version is correct. Do you think I should add that to this description?

src/extension_utils/utils.ts
```diff
@@ -317,53 +353,63 @@ export const promptInstallPythonDependencies = (
...
+
+    return getPathToScript(
+        context,
+        path.join(CONSTANTS.FILESYSTEM.PYTHON_VENV_DIR, subFolder),
+        HELPER_FILES.PYTHON_EXE
+    );
 };
 
-export const installPythonDependencies = async (
+export const installPythonVenv = async (
```
nasadigital, line 379: I am not sure if this is the usual case, but the following happened to me:
1. Extension prompted me to install venv.
2. Accepted, waited for a bit.
3. Running python in venv succesfully.
4. Change the interpreter.
5. Now I can't change back, the python extension can't find the venv.

Do you think we should put the venv somewhere where the python extension can find it more easily?

nasadigital (commented):
The experience was much smoother than I expected! Awesome job :)

locales/en/package.i18n.json
andreamah: Ah okok! I'll try to make it clearer :)

src/extension_utils/utils.ts
andreamah: The venv in the extension backend should only be used in the case where the current venv can't support the extension. So, it should not be in the interpreter select menu and the extension should automatically be able to switch to the custom environment as a fallback when try to select an interpreter that is global + doesn't have all the dependencies.

nasadigital: I see. The extension is correctly finding the venv, I had a different interpreter selected. I think it should be fine!

package.nls.json
andreamah: oops! will fix

src/check_if_venv.py
andreamah: As of right now, I can't really find another better way of doing this. Is the current way fine?

src/extension_utils/utils.ts
```diff
@@ -317,53 +353,63 @@ export const promptInstallPythonDependencies = (
...
                         if (installChoice === DialogResponses.INSTALL_NOW) {
-                            return installPythonDependencies(
-                                context,
-                                pythonExecutable,
-                                pathToLibs
-                            );
+                            return installPythonVenv(context, pythonExecutable);
                         } else {
-                            return false;
+                            return "";
```
isadorasophia, line 362: I'd suggest adding some comments around here so someone who is not familiar with the codebase can quickly understand what's going on -- e.g. why are we returning an empty string here?

src/extension_utils/utils.ts
```diff
@@ -317,53 +353,63 @@ export const promptInstallPythonDependencies = (
...
+
+        console.error(err);
+
+        return pythonExecutable;
+    }
+
+    if (!(await installDependencies(context, pythonPath))) {
+        vscode.window
+            .showErrorMessage(
+                `Dependency download for the custom virtual environment for download could not be completed. Using original interpreter at: ${pythonExecutable}.`,
```
isadorasophia: Should we use a constant error string instead for these error messages?

src/check_if_venv.py
nasadigital: This looks fine to me! I would add a comment to make things more clear, though.

src/extension_utils/utils.ts
```diff
@@ -372,9 +418,194 @@ export const installPythonDependencies = async (
...
+        context,
+        CONSTANTS.FILESYSTEM.OUTPUT_DIRECTORY,
+        HELPER_FILES.CHECK_PYTHON_DEPENDENCIES
+    );
+    try {
+        const { stdout } = await executePythonCommand(
+            pythonPath,
+            `"${dependencyCheckerPath}"`
+        );
+        console.info(stdout);
```
isadorasophia, line 449: As I mentioned above, I think it'd be nice to have a bit more of documentation for these -- just so it's easier to debug in the future. Like explaining why we need to output these for stdout and things like that.

locales/en/package.i18n.json
andreamah: I changed it! Is it better now?

andreamah: description updated for clarity!

src/check_if_venv.py
andreamah: ah ok cool! will add

src/extension_utils/utils.ts
```diff
@@ -372,9 +418,194 @@ export const installPythonDependencies = async (
...
+
         console.error(err);
         this.logToOutputChannel(errorChannel, err.toString(), true /* show */);
-        installed = false;
+        return false;
     }
-    return installed;
+};
+
+export const setupEnv = async (
```
isadorasophia: nit: I would make this method more descriptive somehow.
e.g. you can break it down as `GetPythonExecutableName(...)` and then `SetupEnv(...)`

src/extension_utils/utils.ts
```diff
@@ -262,52 +251,99 @@ export const checkConfig = (configName: string): boolean => {
...
+};
+
+export const executePythonCommand = async (
+    pythonExecutableName: string,
+    command: string
+) => {
+    return exec(`${createEscapedPath(pythonExecutableName)} ${command}`);
+};
+
+export const validPythonVersion = async (pythonExecutableName: string) => {
```
isadorasophia: also nit: you can rename for validatePythonVersion

src/check_if_venv.py
andreamah: added! c:

src/extension_utils/utils.ts
```diff
@@ -372,9 +418,194 @@ export const installPythonDependencies = async (
...
+    if (!(await checkForDependencies(context, pythonExecutableName))) {
+        // environment needs to install dependencies
+        if (!(await checkIfVenv(context, pythonExecutableName))) {
+            pythonExecutableName = await getPythonVenv(context);
+            if (await hasVenv(context)) {
+                // venv in extention exists with wrong dependencies
+                if (
+                    !(await checkForDependencies(context, pythonExecutableName))
+                ) {
+                    await installDependencies(context, pythonExecutableName);
```
isadorasophia: Do we want to do anything if this fails?

src/extension_utils/utils.ts
```diff
@@ -372,9 +418,194 @@ export const installPythonDependencies = async (
...
+                        CONSTANTS.INFO.INSTALL_PYTHON_DEPS,
+                        DialogResponses.INSTALL_NOW,
+                        DialogResponses.DONT_INSTALL
+                    )
+                    .then(
+                        async (
+                            installChoice: vscode.MessageItem | undefined
+                        ) => {
+                            if (installChoice === DialogResponses.INSTALL_NOW) {
+                                await installDependencies(
```
isadorasophia: Same as above, do we want to do anything if this fails?

package.nls.json
```diff
@@ -12,7 +12,7 @@
   "deviceSimulatorExpressExtension.commands.microbit.openSimulator": "[micro:bit] Open Simulator",
   "deviceSimulatorExpressExtension.commands.microbit.newFile": "[micro:bit] New File",
   "deviceSimulatorExpressExtension.configuration.title": "Device Simulator Express configuration",
-  "deviceSimulatorExpressExtension.configuration.properties.configEnvOnChange": "When you change the Python Interpreter, the dependencies for the Device Simulator Express will automatically be configured for the extension dependencies (and the interpreter will auto-change if necessary).",
+  "deviceSimulatorExpressExtension.configuration.properties.configEnvOnChange": "When you change the Python interpreter, the Device Simulator Express will atuomatically configure itself for the required dependencies.",
```
nasadigital: 'automatically' is misspelled.

src/check_if_venv.py
isadorasophia: I'd suggest adding a comment explaining that too!

src/extension_utils/utils.ts
```diff
@@ -164,45 +168,30 @@ export function generateCPXConfig(): void {
...
-};
 
-export const setPythonExectuableName = async () => {
-    // Find our what command is the PATH for python
-    let executableName: string = "";
-    const dependencyCheck = await checkPythonDependency();
-    if (dependencyCheck.installed) {
-        executableName = dependencyCheck.dependency;
-    } else {
+export const checkForPip = async (pythonExecutableName: string) => {
```
isadorasophia: nit: this is all nit suggestions, so don't worry all much about it. I would use the "check" keyword for methods that validate something (throw an exception or show up an error window).

We can be more descriptive for methods that return a boolean (such as this one), e.g. `IsPipInstalled()`

nasadigital (approved):
LGTM. I noticed that the linter outputs a errors before "activating" the extension in the session (before we point to the venv), do you think we can do something about that?

isadorasophia (approved):
I am not entirely familiar with the best practices managing packages for VS extensions, but LGTM otherwise. I mostly made some nit suggestions as I had some trouble understanding some of the details at first :)

package.nls.json
andreamah: > LGTM. I noticed that the linter outputs a errors before "activating" the extension in the session (before we point to the venv), do you think we can do something about that?

For this, the pylint configuration is added upon activation, so it's hard to have it recognize it before activating :(. Based on our current structure, pylint doesn't know to look in our outfolder until we specify.

src/extension_utils/utils.ts
andreamah: Oh right! Thanks for catching that :)

src/extension_utils/utils.ts
andreamah: ah okok! I'll try to add more!

src/extension_utils/utils.ts
andreamah: Since the prompt includes a variable name, should I make part of the string a constant? Or should I leave it be?

isadorasophia: We should be able to declare it as a constant with a paremeter for the variable name ("%1", "%s", or something of the sorts)?

src/extension_utils/utils.ts
andreamah: installDependencies has a try-catch block inside, so it should never throw an error. The only things outside of the try-catch block in installDependencies is getPathToScript (which should never fail) and checkForPip (which also has its own try-catch)

src/extension_utils/utils.ts
andreamah: same as above also ^

src/extension_utils/utils.ts
andreamah: ah ok! I'll make some changes for that

src/extension_utils/utils.ts
isadorasophia: Doesn't it return false if pip is not installed? (E.g. nothing gets installed) I was more thinking on these lines

src/extension_utils/utils.ts
andreamah: @xnkevinnguyen I'm redirecting this question to you! :'''D is this possible?

xnkevinnguyen: I think it's ok to do this for template strings. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals, if you want to do constant, you can make a function that returns the value, but showing error message here should be ok as long as it's consistently done this way

xnkevinnguyen (approved):

src/extension_utils/utils.ts
isadorasophia: Ugh, I couldn't find anything super easy <_< Similar to what Kevin suggested, it should be all right to declare a constant as the string is?
`'Dependency download for the custom virtual environment for download could not be completed. Using original interpreter at: ${pythonExecutable}.'`

isadorasophia: You could also create a wrapper, as it's likely we will use more of that in the future?
https://stackoverflow.com/a/7790858

src/extension_utils/utils.ts
andreamah: fixed in the commit! Will throw an error if dependency download fails :)

src/extension_utils/utils.ts
andreamah: @isadorasophia is that ok? ^

nasadigital:
By the way, does the CI pipeline test the whole flow (from installing the extension to running an example)?

src/extension_utils/utils.ts
andreamah: it can be added as a refactoring task in the future if it's needed! It may too much overhead for this PR :(((((

isadorasophia: No worries! You can just create a work item for that (mostly so we don't forget haha) and deal with it afterwards

PR #223, opened 2020-02-26
Release Note

andreamah:
# Description:

Release note, which only shows once (upon activating after update)

## Type of change

# Limitations:

# Testing:

# Checklist:


src/service/PopupService.ts
```diff
@@ -0,0 +1,16 @@
+// import { Webview } from "vscode";
+import * as vscode from "vscode";
+import { LATEST_RELEASE_NOTE } from "../latest_release_note";
+
+export class PopupService {
+    public static OPEN_RELEASE_NOTE() {
```
nasadigital: In this project function names are written in pascal case, instead of all capital snake case.

src/release_notes/release_note_2020-02-25.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
+          <html lang="en">
+          <head>
+              <meta charset="UTF-8">
+              <meta name="viewport" content="width=device-width, initial-scale=1.0">
+              <title>Cat Coding</title>
```
nasadigital: Is this file supposed to be checked in?

andreamah: ah oops!! I'll git rm it

src/extension.ts
```diff
@@ -119,6 +121,18 @@ export async function activate(context: vscode.ExtensionContext) {
...
+        "release_note_" + getPackageInfo(context).extensionVersion;
+    const viewedReleaseNote = context.globalState.get(
+        currVersionReleaseName,
+        false
+    );
+
+    if (!viewedReleaseNote) {
+        PopupService.openReleaseNote();
+        context.globalState.update(currVersionReleaseName, true);
+    }
```
nasadigital, line 135: My understanding is that every time we update the extension, users who receive the update will be shown `LATEST_RELEASE_NOTE` .
When developers are ready for a new version, they need to NOT forget to update `LATEST_RELEASE_NOTE`. Otherwise on update VS Code will show the old release notes.
I think that adds overhead to updating the extension. What can we do to streamline the process?

andreamah: I still need to look into an HTML loader that can load an HTML file into a string. Then, it might be easier to just select the newest release note and make sure that certain attributes can match the actual version number before loading it in. This is just a quick solution, since we're on a tight deadline :((((
Otherwise, I can rename the variable in latest_release_note.ts to be more unique to the release date so that it's more noticeable which release note it's pointing to! 

What do you think?

nasadigital: I think it's good to go for this release. Thanks for the clarification!

nasadigital (approved):

vandyliu (approved):
awesome :D

PR #228, opened 2020-02-27
Venv fixes

andreamah:
# Description:

User flow fixes, organization, and asks users if they want install globally if they refuse to do it in a venv

## Type of change

# Limitations:

# Testing:

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/src/extension_utils/utils.ts
+++ b/src/extension_utils/utils.ts
@@ -85,7 +85,7 @@ export function tryParseJSON(jsonString: string): any | boolean {
         if (jsonObj && typeof jsonObj === "object") {
             return jsonObj;
         }
-    } catch (exception) { }
+    } catch (exception) {}
 
     return false;
 }
--- a/src/extension_utils/utils.ts
+++ b/src/extension_utils/utils.ts
@@ -565,10 +565,12 @@ export const setupEnv = async (
     let pythonExecutableName = originalPythonExecutableName;
 
     if (!(await areDependenciesInstalled(context, pythonExecutableName))) {
+        const pythonExecutableNameVenv = await getPythonVenv(context);
         // environment needs to install dependencies
         if (!(await checkIfVenv(context, pythonExecutableName))) {
-            const pythonExecutableNameVenv = await getPythonVenv(context);
+            console.log("uuuuwuuu 1")
             if (await hasVenv(context)) {
+                console.log("uuuuwuuu 2")
                 // venv in extention exists with wrong dependencies
                 if (
                     !(await areDependenciesInstalled(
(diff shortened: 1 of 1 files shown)


isadorasophia (approved):

src/extension_utils/utils.ts
```diff
@@ -607,47 +592,38 @@ export const setupEnv = async (
...
+                                    context,
+                                    pythonExecutableName
+                                );
+                            } else {
+                                await vscode.window
+                                    .showInformationMessage(
+                                        CONSTANTS.INFO.ARE_YOU_SURE,
+                                        DialogResponses.INSTALL_NOW,
+                                        DialogResponses.DONT_INSTALL
+                                    )
```
nasadigital, line 605: I think it would be ok to remove this prompt, as it is the third prompt in a row asking about installing dependencies. Up to you.

src/extension_utils/utils.ts
```diff
@@ -607,47 +592,38 @@ export const setupEnv = async (
...
+                                                | undefined
+                                        ) => {
+                                            if (
+                                                installChoice2 ===
+                                                DialogResponses.INSTALL_NOW
+                                            ) {
+                                                await installDependenciesWrapper(
+                                                    context,
+                                                    pythonExecutableName
+                                                );
```
nasadigital, line 619: If you don't install the dependencies (perhaps you miss the prompts) and try running some micro:bit code, VS Code only shows a generic error:
```
[ERROR] Traceback (most recent call last):
  File "c:\Users\user\.vscode\extensions\vscode-python-devicesimulator\out\process_user_code.py", line 26, in <module>
    from common.telemetry import telemetry_py
  File "c:\Users\user\.vscode\extensions\vscode-python-devicesimulator\out\common\telemetry.py", line 1, in <module>
    from applicationinsights import TelemetryClient
ModuleNotFoundError: No module named 'applicationinsights'
```
It would be much more helpful if we could show the user a more helpful message.

nasadigital (approved):
LGTM! Thanks for the fix 🛫 🌕

src/extension_utils/utils.ts
andreamah: True! Perhaps that'd be good for another release, since we're rushing for the release in an hour :'''(

src/extension_utils/utils.ts
andreamah: perhaps we should leave it in for now, since people who just close popups on instinct should be advised that the extension won't work without installing dependencies! Maybe we can incorporate this info into an earlier popup later on :0

nasadigital (commented):
It would also be very useful to add some telemetry to the prompts and see where users end up installing the dependencies to make data driven decisions in the future.

PR #233, opened 2020-03-01
Deploy code to micro:bit and removed need to rename file to deploy to CPX

vandyliu:
# Description:

- Added functionality to deploy code to micro:bit
- Added command for it, which must be enabled by enabling preview mode

## Type of change

# Limitations:

# Testing:

- [ ] Test A
- [ ] Test B

# Checklist:


src/extension.ts
```diff
@@ -829,6 +776,94 @@ export async function activate(context: vscode.ExtensionContext) {
...
+                    outChannel,
+                    CONSTANTS.INFO.DEPLOY_SUCCESS
+                );
+                break;
+
+            case "no-device":
+                telemetryAI.trackFeatureUsage(
+                    errorCommandDeployWithoutDevice
+                );
+                if (device === CONSTANTS.DEVICE_NAME.CPX) {
```
vandyliu, line 786: need to design error message for microbit

src/device.py
```diff
@@ -61,22 +65,53 @@ def find_device_directory(self):
...
+    def deployToMicrobit(self):
+        try:
+            # Temporarily redirecting stdout because there are some print statements in uflash library
+            fake_stdout = open(os.devnull, "w")
+            _stdout = sys.stdout
+            sys.stdout = fake_stdout
+            uflash.flash(path_to_python=self.file_path)
+            sys.stdout = _stdout
+            message = {"type": "complete"}
+        except RuntimeError:
```
nasadigital, line 97: Do we know if this RuntimeError is always caused by a low python version?

src/device.py
```diff
@@ -61,22 +65,53 @@ def find_device_directory(self):
...
+            message = {"type": "wrong-python-version"}
```
nasadigital: _&ast;nit&ast;_
Judging from the message this seems to be _low_ python version. Wrong python version suggests to me that there are particular version that don't work (as opposed to all python versions < 3.3).

src/extension.ts
```diff
@@ -710,7 +710,10 @@ export async function activate(context: vscode.ExtensionContext) {
                 CONSTANTS.ERROR.NO_FILE_TO_RUN,
                 DialogResponses.MESSAGE_UNDERSTOOD
             );
-        } else if (!utils.validCodeFileName(currentFileAbsPath)) {
+        } else if (
+            device == CONSTANTS.DEVICE_NAME.CPX &&
+            !utils.validCodeFileName(currentFileAbsPath)
```
nasadigital: Have you thought about renaming the file in the background and deploying it? That we way we won't have to bother the user to rename it.

nasadigital (commented):
Worked on first try!

src/device.py
vandyliu: Yep :)

src/device.py
vandyliu: Yeah, I agree. I was wondering if we should just remove it though because our extension is promoted as having Python3.7+ as a requirement.

src/extension.ts
vandyliu: Rename the file in the background then rename it back after it is done?

nasadigital: I was thinking more of how in unix you can do copy and rename at the same time:
`$ cp local_disk/python_script.py microbit/code.py`
But whatever you think makes sense should be good. Doesn't have to be in this PR.

vandyliu: I see what you mean now. I'll update it in this PR :D

nasadigital: Sounds good! Do you think there could be some drawbacks to this approach (taking care of the script name in the background rather than explicitly asking the user to rename the file)?

vandyliu: There could be like if someone makes a file 'main.py', deploys it and then makes a file called 'code.py' and then deploys it, there may be problems because it may still be running the wrong file so we could just force it to be named 'code.py' always.

In response to the OG question, I think the pros outweigh the cons especially because I can remove a lot of unnecessary pop ups now.

src/constants.ts
```diff
@@ -85,6 +77,10 @@ export const CONSTANTS = {
             "error.invalidPythonPath",
             'We found that your selected Python interpreter version is too low to run the extension. Please upgrade to version 3.7+ or select a different interpreter (CTRL+SHIFT+P and type "python.selectInterpreter") and restart the application.'
         ),
+        LESS_PYTHON_VERSION_FOR_MICROBIT_DEPLOYMENT: localize(
```
nasadigital: _&ast;nit&ast;_
`LOW` instead of `LESS`?

nasadigital (approved):
LGTM! Great job removing a lot of code 👍

src/constants.ts
vandyliu: will update it

markAtMicrosoft (approved):

PR #240, opened 2020-03-04
Venv Telemetry and Structure Refactoring

andreamah:
# Description:
Adding telemetry to the new venv setup process merged in #218 .

EDIT: also re-structured setup code. Now, the code for setup is abstracted into a SetupService object. The telemetryAI object is now passed in within the SetupService constructor and setupEnv is called whenever the setup process needs to be triggered.

## Type of change

# Testing:

Tested setup process (common flows)
- installing virtual environment in extension
- installing with personal virtual environment

# Checklist:


src/constants.ts
```diff
@@ -374,6 +374,20 @@ export enum TelemetryEventName {
...
     MICROBIT_PERFORMANCE_OPEN_SIMULATOR = "MICROBIT.PERFORMANCE.OPEN.SIMULATOR",
+
+    // Venv options
+    SETUP_VENV_CREATION_ERR = "SETUP.VENV.CREATION.ERR",
+    SETUP_NO_PIP = "SETUP.NO.PIP",
+    SETUP_DEP_INSTALL_FAIL = "SETUP.DEP.INSTALL.FAIL",
+    SETUP_AUTO_RESOLVE_PYTHON_PATH = "SETUP.AUTO.RESOLVE.PYTHON.PATH",
+    SETUP_NO_PYTHON_PATH = "SETUP.NO.PYTHON.PATH",
+    SETUP_DOWNLOAD_PYTHON = "SETUP.DOWNLOAD.PYTHON",
+    SETUP_BAD_PYTHON_PATH = "SETUP.BAD.PYTHON.PATH",
```
xnkevinnguyen: nit:can we change it for wrong python path instead of bad python path

xnkevinnguyen:
Would be nice to have venv as a service where you would only set telemetry ai once as an attribute

src/constants.ts
andreamah: Bad python path pertains to having an interpreter path to a file that does not exist. I will try to change the name to make it make more sense!

src/constants.ts
```diff
@@ -374,6 +374,20 @@ export enum TelemetryEventName {
     MICROBIT_PERFORMANCE_DEPLOY_DEVICE = "MICROBIT.PERFORMANCE.DEPLOY.DEVICE",
     MICROBIT_PERFORMANCE_NEW_FILE = "MICROBIT.PERFORMANCE.NEW.FILE",
     MICROBIT_PERFORMANCE_OPEN_SIMULATOR = "MICROBIT.PERFORMANCE.OPEN.SIMULATOR",
+
+    // Venv options
+    SETUP_VENV_CREATION_ERR = "SETUP.VENV.CREATION.ERR",
```
nasadigital, line 379: Are you familiar with how these strings are used in app insights?

xnkevinnguyen (approved):
Nice work! Will be nice to have telemetry

src/constants.ts
vandyliu: They're just strings. It's easier to search for related events this way and it was also the way that the previous interns did it.

nasadigital: Thanks for clarifying!
I was thinking we could structure them in a more useful way (For e.g. everything related to setup starting with `SETUP.***`, all errors related to set up would be `SETUP.ERROR.***`, as opposed to having the name separated with dots instead of underscores `A_B_C_D` -> `A.B.C.D`) to streamline filtering events. Sayyeda showed me how it all looks in AppInsights and it seems there is no support for this kind of filtering out of the box (unless you write queries), so I think it's fine to keep it this way for now.

nasadigital (approved):
LGTM! 🏖🏖🏖

src/constants.ts
vandyliu: Definitely a good idea

PR #283, opened 2020-03-28
Clue Slideshow Library

andreamah:
# Description:

Implementing backend library for Adafruit Slideshow (front end is already taken care of, since the backend sends frame-by-frame).

**ORIGINAL CODE BY ADAFRUIT:** https://github.com/adafruit/Adafruit_CircuitPython_Slideshow/blob/master/adafruit_slideshow.py
- For the most part, I re-did the library. I did borrow some docstrings and function names. 

**DOCS:** https://circuitpython.readthedocs.io/projects/slideshow/en/latest/api.html

### File breakdown:

**adafruit_slideshow.py**: all of the libraries needed to run a slideshow
--> The **SlideShow** class is the main class containing slideshow model. There are two ways of proceeding: advance() and update(). Advance() will definitely change the active slideshow photo, whereas update() will check the last time you advanced and will only advance if `SlideShow.dwell` seconds has passed. If auto_advance is False, update() will never work.
--> **self.BASE_DWELL** and **self.BASE_DWELL_DARK** exist as a bit of a buffer to add when the photos are fully shown and when the screen is "dark" (applicable during fade effect) respectively. Using no delay on top of the self.dwell value seemed to have made the animation too fast

**test_adafruit_slideshow.py**: tests for slideshow
--> only iterates forwards and backwards on the photos using advance()

### Extra features:
Since this library supports relative file locations, I needed to get the address to the users's file. This was already done in the CPX code, so I modified it so that this was stored in common instead. Thus, both CPX and CLUE could access this value. 
 - This explains the changes to utils.py, express.py, and process_user_code.py

## Type of change

# Limitations:

# Testing:

Below are some tests that I ran. The "img" folder contained ~6 images that fell into one of these categories for height and width:
- 240 by 240 exactly
- more than 240 in height and width
- less than 240 in height and width
- more that 240 in width but less than 240 in height

I also tested with invalid .bmp images and .png images to make sure that they were properly skipped.

```
from adafruit_slideshow import SlideShow, PlayBackDirection, PlayBackOrder
import board

# Create the slideshow object that plays through once alphabetically.
slideshow = SlideShow(
    board.DISPLAY,
    dwell=1,
    folder="img",
    loop=True,
    fade_effect=True,
    auto_advance=True,
    order=PlayBackOrder.RANDOM,
    direction=PlayBackDirection.BACKWARD,
)
for i in range(8):
    slideshow.advance()
```

```
from adafruit_slideshow import SlideShow, PlayBackDirection, PlayBackOrder
import board

# Create the slideshow object that plays through once alphabetically.
slideshow = SlideShow(
    board.DISPLAY,
    dwell=3,
    folder="img",
    loop=True,
    fade_effect=False,
    auto_advance=True,
    order=PlayBackOrder.ALPHABETICAL,
    direction=PlayBackDirection.FORWARD,
)
while slideshow.update():
    pass
```

```
from adafruit_slideshow import SlideShow, PlayBackDirection, PlayBackOrder
import board

# Create the slideshow object that plays through once alphabetically.
slideshow = SlideShow(
    board.DISPLAY,
    dwell=1,
    folder="img",
    loop=True,
    fade_effect=False,
    auto_advance=False,
    order=PlayBackOrder.ALPHABETICAL,
    direction=PlayBackDirection.FORWARD,
)

for i in range(2):
    slideshow.advance()
while slideshow.update():
    pass
```

```
from adafruit_slideshow import SlideShow, PlayBackDirection, PlayBackOrder
import board

# Create the slideshow object that plays through once alphabetically.
slideshow = SlideShow(
    board.DISPLAY,
    dwell=1,
    folder="img",
    loop=True,
    fade_effect=False,
    auto_advance=True,
    order=PlayBackOrder.ALPHABETICAL,
    direction=PlayBackDirection.FORWARD,
)

for i in range(2):
    slideshow.advance()
while slideshow.update():
    pass
```

# Checklist:


src/clue/adafruit_slideshow.py
```diff
@@ -0,0 +1,303 @@
...
+import os
+import base64
+from io import BytesIO
+from base_circuitpython import base_cp_constants as CONSTANTS
+import time
+import collections
+from random import shuffle
+from common import utils
+
+# taken from adafruit
```
isadorasophia, line 12: Add path to file on the GitHub repo? Might be useful later on

src/process_user_code.py
```diff
@@ -100,7 +100,7 @@ def handle_user_prints():
 
 # Execute User Code Thread
 def execute_user_code(abs_path_to_code_file):
-    cpx._Express__abs_path_to_code_file = abs_path_to_code_file
```
isadorasophia, line 103: Do we still need this cpx field? (It looks like we use it at `debug_user_code.py`)

src/clue/adafruit_slideshow.py
```diff
@@ -0,0 +1,303 @@
...
+
+        self._curr_img_handle = new_img
+        self._curr_img = new_path
+        self._img_start = time.monotonic()
+        return True
+
+    def _send(self, img):
+        # sends current bmp_img to the frontend
+        buffered = BytesIO()
+        img.save(buffered, format="BMP")
```
isadorasophia: Should we create const values for these "BMP" and ".bmp"? In case we change that in the future.

isadorasophia (approved):
Thanks for the PR description!! 💛 LGTM

src/process_user_code.py
andreamah: True! I'll fix that in debug_user_code :)

src/clue/adafruit_slideshow.py
```diff
@@ -0,0 +1,312 @@
...
+        """Enable auto-advance based on dwell time.  Set to ``False`` to manually control."""
+
+        self.loop = loop
+        """Specifies whether to loop through the images continuously or play through the list once.
+        ``True`` will continue to loop, ``False`` will play only once."""
+
+        self.fade_frames = 8
+        """Whether to include the fade effect between images. ``True`` tells the code to fade the
+           backlight up and down between image display transitions. ``False`` maintains max
+           brightness on the backlight between image transitions."""
```
vandyliu: Kind of confused with the value and the description. Value is 8 but description makes it seem like it is just a boolean.

src/clue/adafruit_slideshow.py
```diff
@@ -0,0 +1,312 @@
...
+                new_img = Image.open(new_path)
+
+                new_img = new_img.convert("RGBA")
+                new_img.putalpha(255)
+
+                new_img = new_img.crop(
+                    (0, 0, CONSTANTS.SCREEN_HEIGHT_WIDTH, CONSTANTS.SCREEN_HEIGHT_WIDTH)
+                )
+
+                if new_img.size[0] < 240 or new_img.size[1] < 240:
```
vandyliu, line 216: should we use the constants instead of 240?

vandyliu (approved):
Good stuff

src/clue/adafruit_slideshow.py
andreamah: yes!

src/clue/adafruit_slideshow.py
andreamah: oops, I put the doctstring in the wrong place!

PR #286, opened 2020-03-31
Buttons Integration for clue

xnkevinnguyen:
# Description:
Both the backend and front end for buttons on clue [AB#34363](https://microsoftgarage.visualstudio.com/002806e2-ebaa-4672-9d2e-5fe5d29154ef/_workitems/edit/34363)

cluebuttons

## Type of change

# Limitations:

# Testing:

- [ ] Test A
- [ ] Test B

# Checklist:


src/clue/adafruit_clue.py
```diff
@@ -199,15 +199,65 @@ class Clue:  # pylint: disable=too-many-instance-attributes, too-many-public-met
...
+          while True:
+              if clue.button_b:
+                  print("Button B pressed")
+        """
+        return self._b
+
+    def __update_button(self, button, value):
+        if button == "button_a":
+            if value:
+                self.__pressed_buttons.add("A")
```
andreamah, line 238: With this, it might be better to have a dict with each button mapped to whether they are pressed. This way, it might be a little clearer. 
Something like:
`pressed_btns["A"] = value or pressed_btns["A"]`

src/clue/adafruit_clue.py
```diff
@@ -199,15 +199,65 @@ class Clue:  # pylint: disable=too-many-instance-attributes, too-many-public-met
...
+    @property
+    def were_pressed(self):
+        """Returns a set of the buttons that have been pressed.
+        To use with the CLUE:
+        .. code-block:: python
+          from adafruit_clue import clue
+          while True:
+              print(clue.were_pressed)
+        """
+        ret = self.__pressed_buttons.copy()
```
andreamah, line 254: I see here that this is probably why you opted for a set. But, it might be more worth it to use a dict for clarity (?)

src/view/components/microbit/Microbit_svg.tsx
```diff
@@ -4,6 +4,8 @@
 // Adapted from : https://makecode.microbit.org/#editor
 
 import * as React from "react";
+import "../../styles/SimulatorSvg.css";
```
andreamah, line 7: Does SimulatorSvg also work for CPX? If not, maybe a different name would be clearer?

andreamah (changes requested):
Amazing! That was so fast and well done!! 🐰

src/view/components/microbit/Microbit_svg.tsx
xnkevinnguyen: At the moment, this css is shared amongst micro:bit and Clue. The goal is that next interns will be able to use this css for future simulators svgs as well. If we were to refactor the cpx, we should use this css as well.

nasadigital (approved):
Great job, LGTM! 💯

src/clue/adafruit_clue.py
nasadigital: If we use dictionary, it makes it seem that we are mapping strings to other objects, but in this case (if I understand correctly) we only map to boolean (true/false).
What do you think?

vandyliu: I used a set because that is how it is returned in the actual CLUE

src/clue/adafruit_clue.py
vandyliu: Same as below, I implemented it so that it works the same as the CLUE library.
Here is the original:
```
        ret = set()
        pressed = self._gamepad.get_pressed()
        for button, mask in (("A", 0x01), ("B", 0x02)):
            if mask & pressed:
                ret.add(button)
        return ret
```

andreamah (approved):
Amazing! Merge it! 🙂

PR #289, opened 2020-04-01
Sensors for clue

xnkevinnguyen:
# Description:
Make all the sensors for clue. 
[AB#34292](https://microsoftgarage.visualstudio.com/002806e2-ebaa-4672-9d2e-5fe5d29154ef/_workitems/edit/34292)
[AB#34296](https://microsoftgarage.visualstudio.com/002806e2-ebaa-4672-9d2e-5fe5d29154ef/_workitems/edit/34296)
[AB#34297](https://microsoftgarage.visualstudio.com/002806e2-ebaa-4672-9d2e-5fe5d29154ef/_workitems/edit/34297)
 List of sensors: Temperature, Light accelerometer, Leds, push_button, gpio, speaker, sound , pressure, humidity, gesture, proximity
## Type of change

**Testing Steps**
1. Run the code to display sensor values 
```

clue_data = clue.simple_text_display(text_scale=2)

while True:
    clue_data[0].text = "Acceleration: {:.2f} {:.2f} {:.2f} m/s^2".format(
        *clue.acceleration
    )
    clue_data[1].text = "Pressure: {:.3f} hPa".format(clue.pressure)
    clue_data[2].text = "Altitude: {:.1f} m".format(clue.altitude)
    clue_data[3].text = "Temperature: {:.1f} C".format(clue.temperature)
    clue_data[4].text = "Humidity: {:.1f} %".format(clue.humidity)
    clue_data[5].text = "Proximity: {}".format(clue.proximity)
    clue_data[6].text = "Gesture: {}".format(clue.gesture)
    clue_data[7].text = "Color: R: {} G: {} B: {} C: {}".format(*clue.color)
    clue_data[8].text = "Button A: {}".format(clue.button_a)
    clue_data[9].text = "Button B: {}".format(clue.button_b)

    clue_data.show()
```

2. Modify sensors with Value to display the values on the clue screen. Gesture should return an integer depending on it's mapped value (ex: UP=1)


nasadigital:
Do you think you can take a look at some of @andreamah's PRs (E.g. #283, #290) and add more testing instructions?

src/clue/adafruit_clue.py
```diff
@@ -334,13 +721,23 @@ def simple_text_display(
...
-            self.__update_button(button_name, new_state.get(button_name))
+    def __update_button(self, button, value):
+        if button == CONSTANTS.CLUE_STATE.BUTTON_A:
+            if value:
+                self.__state[CONSTANTS.CLUE_STATE.PRESSED_BUTTONS].add("A")
+            self.__state[CONSTANTS.CLUE_STATE.BUTTON_A] = value
+        elif button == CONSTANTS.CLUE_STATE.BUTTON_B:
+            if value:
+                self.__state[CONSTANTS.CLUE_STATE.PRESSED_BUTTONS].add("B")
+            self.__state[CONSTANTS.CLUE_STATE.BUTTON_B] = value
```
nasadigital: You can streamline this function by having something like:
`self.__state[button] = value`
at the start (or end) and cleaning up the rest of the conditionals.

src/clue/test/test_adafruit_clue.py
```diff
@@ -62,18 +62,142 @@ def _send_helper(self, image):
...
         assert not clue.button_b
 
         assert set(["A", "B"]) == clue.were_pressed
         assert set() == clue.were_pressed
+
+    def test_acceleration(self):
+        MOCK_MOTION_X_A = 1
+        MOCK_MOTION_Y = 2
+        MOCK_MOTION_Z = 3
+        MOCK_MOTION_X_B = 4
```
nasadigital: Why did you decide on this approach as opposed to using `pytest.parametrize`?

xnkevinnguyen:
> Do you think you can take a look at some of @andreamah's PRs (E.g. #283, #290) and add more testing instructions?

Of course! Will take consideration for all my new prs

src/view/components/toolbar/motion/Gesture.tsx
```diff
@@ -0,0 +1,71 @@
...
+            }
+        }
+    };
+
+    private handleOnKeyUp = (
+        e: React.KeyboardEvent,
+        onSendGesture?: (isActive: boolean) => void
+    ) => {
+        if (e.key === CONSTANTS.KEYBOARD_KEYS.ENTER) {
+            this.sensorButtonRef!.current!.setButtonClass(false);
```
nasadigital, line 64: I've never seen this, what does the `!.` operator do?

nasadigital:
differentColors
I'm not sure if you can notice, but there is no gradation inside the red box.

nasadigital (approved):
Great job folks! LGTM 💯 💯 💯

src/view/components/toolbar/motion/Gesture.tsx
xnkevinnguyen: Hey David, this indicated that the value is defined. The reason why it's used is because it's set directly here at line 31 of the same file
                    <SensorButton
                    ref={this.sensorButtonRef}

nasadigital: My bad, I don't quite get it. Do you have any reference?

xnkevinnguyen: A new ! post-fix expression operator may be used to assert that its operand is non-null and non-undefined in contexts where the type checker is unable to conclude that fact. Specifically, the operation x! produces a value of the type of x with null and undefined excluded. Similar to type assertions of the forms <T>x and x as T, the ! non-null assertion operator is simply removed in the emitted JavaScript code. 

https://stackoverflow.com/questions/42273853/in-typescript-what-is-the-exclamation-mark-bang-operator-when-dereferenci

nasadigital: Thanks for clarifying!

xnkevinnguyen:
> differentColors
> I'm not sure if you can notice, but there is no gradation inside the red box.

@nasadigital made it consistent with latest commit 
image

src/clue/test/test_adafruit_clue.py
vandyliu: It slipped my mind 😭 I updated it to use it now. I love pytest.parametrize :D

src/clue/adafruit_clue.py
vandyliu: Updated!

nasadigital (approved):
Thank you both for addressing the comments!

src/service/telemetryHandlerService.ts
```diff
@@ -244,8 +244,8 @@ export class TelemetryHandlerService {
                 break;
         }
         return {
-            deployTelemetryEvent: deployTelemetryEvent,
-            deployPerformanceTelemetryEvent: deployPerformanceTelemetryEvent,
+            deployTelemetryEvent,
```
vandyliu, line 247: Just wondering, is this still accessed the same way?

xnkevinnguyen: Yes, this typing is not necessary. Change was made by linting.

vandyliu: fascinating, didn't know it worked like that. thanks :D

src/view/translations/en.json
```diff
@@ -49,5 +49,54 @@
     "toolbar-microbit-gpio.tryItDescription": "If you would like to see this sensor supported, please +1 the feature addition issue on GitHub! For now, you can try it on MakeCode!",
     "toolbar-microbit-wireless.title": "Bluetooth  & Radio",
     "toolbar-microbit-wireless.description": "micro:bits can use radio waves and bluetooth services to communicate with each other",
```
vandyliu: Can we add period to end of sentence
```suggestion
    "toolbar-microbit-wireless.description": "micro:bits can use radio waves and bluetooth services to communicate with each other.",

```

src/view/translations/en.json
```diff
@@ -49,5 +49,54 @@
     "toolbar-microbit-gpio.tryItDescription": "If you would like to see this sensor supported, please +1 the feature addition issue on GitHub! For now, you can try it on MakeCode!",
     "toolbar-microbit-wireless.title": "Bluetooth  & Radio",
     "toolbar-microbit-wireless.description": "micro:bits can use radio waves and bluetooth services to communicate with each other",
-    "toolbar-microbit-wireless.tryItDescription": "If you would like to see this sensor supported, please +1 the feature addition issue on GitHub! For now, you can try it on MakeCode!"
+    "toolbar-microbit-wireless.tryItDescription": "If you would like to see this sensor supported, please +1 the feature addition issue on GitHub! For now, you can try it on MakeCode!",
+    "toolbar-microbit-gesture-sensor.title": "Gesture",
+    "toolbar-microbit-gesture-sensor.description": "If you move your BBC micro:bit in a certain way (as a gesture) then micro:bit is able to detect this. micro:bit is able to recognise the following gestures: up, down, left, right, face up, face down, freefall, 3g, 6g, 8g, shake.)",
```
vandyliu: ``
```suggestion
    "toolbar-microbit-gesture-sensor.description": "If you move your BBC micro:bit in a certain way (as a gesture) then the micro:bit is able to detect this. It can recognize the following gestures: up, down, left, right, face up, face down, freefall, 3g, 6g, 8g, and shake.",

```

src/view/translations/en.json
```diff
@@ -49,5 +49,54 @@
...
+    "toolbar-clue-temperature-sensor.description": "CLUE uses the BMP280 sensor, an environmental sensor with temperature.This precision sensor from Bosch is the best low-cost, precision sensing solution for measuring temperature with ±1.0°C accuracy.",
+    "toolbar-clue-temperature-sensor.tryItDescription": "You can set the temperature range from your code!",
+    "toolbar-clue-light-sensor.title": "Light/Color Sensor",
+    "toolbar-clue-light-sensor.description": "The light sensor uses APDS9960 and is able to return the red, blue, green and clear light.",
+    "toolbar-clue-light-sensor.tryItDescription": "You can set the color values of the light detected.",
+    "toolbar-clue-accelerometer-sensor.title": "Accelerometer",
+    "toolbar-clue-accelerometer-sensor.description": " The 3-axis accelerometer, can tell you which direction is down towards the Earth (by measuring gravity) or how fast the board is accelerating in 3D space.",
+    "toolbar-clue-accelerometer-sensor.tryItDescription": "Set the acceleration with the sliders.",
+    "toolbar-clue-led.title": "Neopixel",
+    "toolbar-clue-led.description": "The CLUE has one full RGB LED on it's back",
```
vandyliu:
```suggestion
    "toolbar-clue-led.description": "The CLUE has one full RGB LED (NeoPixel) on the back of the device",

```

src/view/translations/en.json
```diff
@@ -49,5 +49,54 @@
...
+    "toolbar-microbit-gesture-sensor.tryItDescription": "Select a gesture and send it by clicking the button.",
+    "toolbar-microbit-compass-sensor.title": "Compass",
+    "toolbar-microbit-compass-sensor.description": "A digital compass is an input sensor that detects magnetic fields. Your BBC micro:bit has an inbuilt compass that can detect the direction in which it is facing.",
+    "toolbar-microbit-compass-sensor.tryItDescription": "If you would like to see this sensor supported, please +1 the feature addition issue on GitHub!",
+    "toolbar-clue-temperature-sensor.title": "Temperature",
+    "toolbar-clue-temperature-sensor.description": "CLUE uses the BMP280 sensor, an environmental sensor with temperature.This precision sensor from Bosch is the best low-cost, precision sensing solution for measuring temperature with ±1.0°C accuracy.",
+    "toolbar-clue-temperature-sensor.tryItDescription": "You can set the temperature range from your code!",
```
vandyliu:
```suggestion
    "toolbar-clue-temperature-sensor.tryItDescription": "You can simulate the temperature in your code!",

```
Range makes it seem like values from [a, b] when it is just one value.

src/view/translations/en.json
```diff
@@ -49,5 +49,54 @@
...
+    "toolbar-microbit-gesture-sensor.description": "If you move your BBC micro:bit in a certain way (as a gesture) then micro:bit is able to detect this. micro:bit is able to recognise the following gestures: up, down, left, right, face up, face down, freefall, 3g, 6g, 8g, shake.)",
+    "toolbar-microbit-gesture-sensor.tryItDescription": "Select a gesture and send it by clicking the button.",
+    "toolbar-microbit-compass-sensor.title": "Compass",
+    "toolbar-microbit-compass-sensor.description": "A digital compass is an input sensor that detects magnetic fields. Your BBC micro:bit has an inbuilt compass that can detect the direction in which it is facing.",
+    "toolbar-microbit-compass-sensor.tryItDescription": "If you would like to see this sensor supported, please +1 the feature addition issue on GitHub!",
+    "toolbar-clue-temperature-sensor.title": "Temperature",
+    "toolbar-clue-temperature-sensor.description": "CLUE uses the BMP280 sensor, an environmental sensor with temperature.This precision sensor from Bosch is the best low-cost, precision sensing solution for measuring temperature with ±1.0°C accuracy.",
```
vandyliu:
```suggestion
    "toolbar-clue-temperature-sensor.description": "CLUE uses the BMP280 sensor, an environmental sensor with temperature. This precision sensor from Bosch is the best low-cost, precision sensing solution for measuring temperature with ±1.0°C accuracy.",

```

src/view/translations/en.json
```diff
@@ -49,5 +49,54 @@
...
+    "toolbar-clue-gpio.title": "Input/Output (GPIO)",
+    "toolbar-clue-gpio.description": "On the bottom edge of your CLUE, you have 25 pins. These pins allow you to really get creative. You can create circuits, connect external things like buzzers and motors and make your own fun projects.",
+    "toolbar-clue-gpio.tryItDescription": "If you would like to see this sensor supported, please +1 the feature addition issue on GitHub!",
+    "toolbar-clue-sound-sensor.title": "Sound",
+    "toolbar-clue-sound-sensor.description": "The CLUE uses a PDM MEMS microphone. PDM is a little like 1-bit PWM. You clock the mic with a 1 MHz - 3 MHz clock rate, and on the data line you'll get a square wave out that syncs with the clock.",
+    "toolbar-clue-sound-sensor.tryItDescription": "If you would like to see this sensor supported, please +1 the feature addition issue on GitHub!",
+    "toolbar-clue-pressure-sensor.title": "Pressure",
+    "toolbar-clue-pressure-sensor.description": "CLUE uses the BMP280 sensor for barometric pressure. Because pressure changes with altitude, and the pressure measurements are so good, you can also use it as an altimeter with  ±1 meter accuracy.",
```
vandyliu:
```suggestion
    "toolbar-clue-pressure-sensor.description": "CLUE uses the BMP280 sensor for barometric pressure. Because pressure changes with altitude, you can also use it as an altimeter with ±1 meter accuracy.",

```
I don't think having it say "so good" is smart without any evidence. I prefer it like ^

src/view/translations/en.json
```diff
@@ -49,5 +49,54 @@
...
+    "toolbar-clue-pressure-sensor.tryItDescription": "You can set the pressure here in hectoPascals (hPa).",
+    "toolbar-clue-humidity-sensor.title": "Humidity",
+    "toolbar-clue-humidity-sensor.description": "The SHT30-D sensor has an excellent ±2% relative humidity and ±0.5°C accuracy for most uses.",
+    "toolbar-clue-humidity-sensor.tryItDescription": "You can set the humidity in percentage (%) here.",
+    "toolbar-clue-gesture-sensor.title": "Gesture",
+    "toolbar-clue-gesture-sensor.description": "The gesture sensor can detect directional gestures (left to right, right to left, up to down, down to up), but in theory more complicated gestures like zig-zag, clockwise or counterclockwise circle, near to far, etc. could also be detected with additional code. ",
+    "toolbar-clue-gesture-sensor.tryItDescription": "Select a gesture and send it by clicking the button.",
+    "toolbar-clue-proximity-sensor.title": "Proximity",
+    "toolbar-clue-proximity-sensor.description": "The proximity sensor uses APDS9960 and is able to return how close an object is to the front of the sensor.",
```
vandyliu:
```suggestion
    "toolbar-clue-proximity-sensor.description": "The proximity sensor uses APDS9960 and is able to return how close an object is to the front of the sensor. This is a number from 0 to 255 where the higher the number the closer an object is to the sensor. You won't be able to translate this into an absolute value in units like inches or millimeters, you can only see how it changes relative to other values",

```

src/view/translations/en.json
```diff
@@ -49,5 +49,54 @@
...
+    "toolbar-clue-proximity-sensor.tryItDescription": "Set relative proximity to the sensor in values from 0 - 255",
```
vandyliu:
```suggestion
    "toolbar-clue-proximity-sensor.tryItDescription": "Set relative proximity to the sensor in values from 0 - 255, where 0 means very close and 255 means very far.",

```

vandyliu (changes requested):
Just some suggestions for the descriptions but other than that it LGTM!

src/view/translations/en.json
xnkevinnguyen: Okok, it was copied from the adafruit website ahaha

xnkevinnguyen:
@vandyliu  Thanks for your text suggestions. They were all applied!!! :)

vandyliu (approved):
SHIP ITTTTT

src/clue/adafruit_clue.py
```diff
@@ -250,10 +265,186 @@ def were_pressed(self):
...
+                                requires a minimum value of 10. 10 is the total
+                                acceleration if the board is not moving, therefore
+                                anything less than 10 will erroneously report a constant
+                                shake detected. (Default 30)
+        :param avg_count: The number of readings taken and used for the average
+                          acceleration. (Default 10)
+        :param total_delay: The total time in seconds it takes to obtain avg_count
+                            readings from acceleration. (Default 0.1)
+        """
+        is_shaken = self.__state[CONSTANTS.CLUE_STATE.GESTURE] == "shake"
```
andreamah: should "shake" be a const?

src/clue/adafruit_clue.py
```diff
@@ -250,10 +265,186 @@ def were_pressed(self):
...
+            from adafruit_clue import clue
+            clue.sea_level_pressure = 1015
+            print("Altitude: {:.1f}m".format(clue.altitude))
+        """
+        altitude = 44330 * (
+            1.0
+            - math.pow(
+                self.__state[CONSTANTS.CLUE_STATE.PRESSURE]
+                / self.__state[CONSTANTS.CLUE_STATE.SEA_LEVEL_PRESSURE],
+                0.1903,
```
andreamah: Should the numbers here be consts also?

src/clue/adafruit_clue.py
```diff
@@ -334,13 +721,24 @@ def simple_text_display(
...
+                    self.__state[event] = new_state.get(event)
 
     # helpers
-    def __update_buttons(self, new_state):
-        # get button pushes
-        for button_name in CONSTANTS.EXPECTED_INPUT_BUTTONS:
-            self.__update_button(button_name, new_state.get(button_name))
+    def __update_button(self, button, value):
+        buttonMapping = {
+            CONSTANTS.CLUE_STATE.BUTTON_A: "A",
```
andreamah: I believe that this same button mapping in display.py for micro:bit was just a class field so that it would be instantiated once instead of after every change. Should this be changed to match that?

andreamah (approved):
Amazing work on all ends!!! 🤩

src/clue/adafruit_clue.py
vandyliu: done!

src/clue/adafruit_clue.py
vandyliu: For me, I don't think it makes sense to make them constants because we don't use them anywhere else and constants may make it more confusing because it's hard to name and in the formulas online they aren't given names? Might just be me.

src/clue/adafruit_clue.py
vandyliu: Done!

src/clue/adafruit_clue.py
xnkevinnguyen: I agree with Andrea here. It's better to not have magic number, if you can find that constant name, it's better if you declare it. If it's used only here, it can be local

xnkevinnguyen: Actually since they don't provide the constants name in the source code, I think it's not worth the additional effort. Linking the source code might help if it's not already done.

vandyliu: Updated it to have constants :p

PR #292, opened 2020-04-02
Microbit Debugger Fix

andreamah:
# Description:

Currently, the microbit does not respond to sensor/button input in debug mode. 

This is because the handle to the microbit model instance is different in `debugger_communication_client.py` compared to everywhere else. Since `micropython `and `microbit `are both packages (although `microbit `is inside of the `micropython `package), instances imported as `from microbit.__model.microbit_model import __mb as mb` and `from micropython.microbit.__model.microbit_model import __mb as mb` deal with two different instances of `mb`. As a result, the `mb `that was getting an updated state was different than the one being referenced by the user code.

## Type of change

# Testing:

Simple code using any sensors/buttons in DEBUG MODE such as this:

```
from microbit import *

while True:
    if button_a.is_pressed():
        display.show(Image.HAPPY)
    else:
        display.show(Image.HEART)

```
# Checklist:


nasadigital (approved):
Great catch! LGTM!

xnkevinnguyen (approved):
LGTM!  Thanks for fixing it

src/common/test/test_debugger_communication_client.py
```diff
@@ -3,10 +3,25 @@
...
-from adafruit_circuitplayground import express
-from common import debugger_communication_client
 from common import constants as CONSTANTS
+
+abs_path_to_parent_dir = os.path.join(
+    os.path.dirname(os.path.abspath(__file__)), "..", ".."
+)
+
+sys.path.insert(0, abs_path_to_parent_dir)
+# Insert absolute path to Micropython libraries for micro:bit into sys.path
```
nasadigital, line 16: I would put this comment above the line it's referring to (or maybe remove the line break on line 20).
Now it seems as if it's referring to the next three lines where we calculate the path.

src/common/test/test_debugger_communication_client.py
```diff
@@ -3,10 +3,25 @@
 from unittest import mock
 import socketio
 import threading
+import os
+import sys
 
-from adafruit_circuitplayground import express
-from common import debugger_communication_client
```
nasadigital, line 8: Just curious, if we used relative paths would the tests use the correct library?

PR #294, opened 2020-04-02
Clue Debugger

andreamah:
# Description:

#### Fixes:
Made debug_mode a field in the common lib rather than just instantiated separately in each microcontroller model. Then, in order to access it, you just need to import utils.

#### New:
Debugging for screen and buttons

### Frontend changes:
Using the debugger adapter, we will freeze the UI to follow proper debugging behaviour since the board cannot read inputs while on PAUSE state.

-The buttons will not be clickable
-The sensors will not send new values
-Gesture button will not send anything


src/debug_user_code.py
```diff
@@ -26,15 +26,17 @@
 )
 sys.path.insert(0, abs_path_to_micropython_lib)
 
+# Insert absolute path to library for CLUE into sys.path
+sys.path.insert(0, os.path.join(abs_path_to_parent_dir, CONSTANTS.CLUE))
+
+# Insert absolute path to Circuitpython libraries for CLUE into sys.path
+sys.path.insert(0, os.path.join(abs_path_to_parent_dir, CONSTANTS.CIRCUITPYTHON))
+
```
nasadigital, line 34: Why do we need to do this?

nasadigital (approved):
Great job on making this work 🌻 🌻 🌻 
LGTM!

src/debug_user_code.py
andreamah: When you run tests, it doesn't run process_user_code.py, which is where this usually happens. This adds src/clue and src/micropython to the sys path so that we can import microbit and clue.

We need to put it in debug_user_code.py rather than its test because some tests import debug_user_code and need these sys.path add-ons in order to import the file.

nasadigital: Thanks for clarifying!
And why do we need this in `process_user_code.py`?

andreamah: The code automatically looks for packages in the out/* directory, but doesn't recursively search. This is so that it can find packages that don't necessarily sit right below out/*. In the case of microbit and clue, they both have their packages defined in sub-folders.

nasadigital: Which modules do we want to use? Why can't use the import statement to do that?

andreamah: for the case of process_user_code.py, we wanted the user to import the packages using a specific way, so we're doing it this way. For all of the imports, we want to keep the method consistent because `import micropython.microbit` and `import microbit` imports two different instances of the microbit package.

nasadigital: Sorry, I am struggling to understand what we are trying to achieve here.
If `import micropython.microbit` and `import microbit` import two different instances, why don't we use one or the other? In which case do we want to use one over the other?

PR #302, opened 2020-04-08
Make clue larger

xnkevinnguyen:
# Description:

Since this was stated during usability testing, the current Clue device is too small. We have increased the size of the clue and the relative size of display to device ratio.

image


nasadigital (approved):
Much easier to see 👓 
LGTM

PR #290, opened 2020-04-01
Clue Terminal and Context Fixes

andreamah:
# Description:

Hi everyone! :) 

### New features:
`_ClueSimpleTextDisplay.show_terminal()` now works! Since its original purpose was to reference to a general terminal built in the firmware, I put the new files in `base_circuitpython`. I also added code in `process_user_code.py `that shows the default opening & closing print statements. This file also takes care of capturing print statements and sending them to the clue terminal.

### Bug fixes:
- After looking more into clue behavior, it turns out that I misunderstood the `board.DISPLAY.show(group)` call. I initially thought that it was called to show the group, but actually, this is a command to "context-switch" to show a specific group. **Only one group can be shown at once, and `board.DISPLAY.show(curr_group)` will change the group on the screen to curr_group.** Any further modifications to the active group will be actively changed on the screen. If modifications are made to the group before calling `board.DISPLAY.show()`, all changes should be present when` board.DISPLAY.show()` is called. 
  - I added support for this within board.py. 
- The screen wasn't properly refreshing (partially due to the lapse in understanding above)
  - `group.py` now always starts with an all-black image every time draw() is called.

### File notes:
- **`base_circuitpython/terminal_handler.py`**: the custom-made terminal class that:
  - always handles the current strings, even when the terminal is not active
  - shows the terminal updating if the active group.
- **`base_circuitpython/tests/terminal_handler_test.py`**: tests for terminal_handler.py
- **`board.py`**: not new, but has major changes. It now keeps track of the active group. Groups will check with this active_group value before sending anything to the frontend.
 
## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

# Testing:

I made a mega-test that goes as follows:

**NOTE: this test needs some slideshow bmp photos in the current working directory!**
```
from adafruit_bitmap_font import bitmap_font

from adafruit_clue import clue
from datetime import datetime
import time

from adafruit_slideshow import SlideShow, PlayBackDirection, PlayBackOrder
import board

# Create the slideshow object that plays through once alphabetically.
slideshow = SlideShow(
    board.DISPLAY,
    dwell=1,
    folder=".",
    loop=True,
    fade_effect=False,
    auto_advance=True,
    order=PlayBackOrder.ALPHABETICAL,
    direction=PlayBackDirection.FORWARD,
)

for i in range(2):
    slideshow.advance()

[...]

PR #304, opened 2020-04-08
Clue Unix Fix

andreamah:
# Description:

Clue simulator should work on Ubuntu with these changes 🛫🚀

### Issue:
On UNIX systems,  the large base64 image string would be split up when being sent to the front-end. To fix this, the front-end will collect incomplete messages and add them to the beginning of  incoming new messages to see if they work there.

For this to work, it relies on the fact that the child process triggered by `process_user_code.py` only sends JSON-formatted strings to stdout. Fortunately, `process_user_code.py` already re-formats print strings in JSON before sending them to the frontend.

## Type of change

# Limitations:

# Testing:

Used mega-test from #290 , but I ran on Ubuntu.
**NOTE: this test needs some slideshow bmp photos in the current working directory!**
```

from datetime import datetime
import time

from adafruit_slideshow import SlideShow, PlayBackDirection, PlayBackOrder
import board

# Create the slideshow object that plays through once alphabetically.
slideshow = SlideShow(
    board.DISPLAY,
    dwell=1,
    folder=".",
    loop=True,
    fade_effect=False,
    auto_advance=True,
    order=PlayBackOrder.ALPHABETICAL,
    direction=PlayBackDirection.FORWARD,
)

for i in range(2):
    slideshow.advance()
    

clue_data = clue.simple_text_display(title="Clue Sensor Data!", title_scale=2,)

val = 100
for i in range(12):
    val += 1
    clue_data.show_terminal()
    time.sleep(0.5)
    print(val)
    print(f"time {datetime.utcnow().strftime('%Y-%m-%d %H:%M:%S.%f')[:-3]}")

    print(clue_data.text_group.hidden)
for i in range(10):
    val += 1
    clue_data[0].text = "Accelerometer:"
    clue_data[1].text = "Gyro:"
    clue_data[2].text = "Magnetic:"
    clue_data[3].text = "Pressure: {:.3f} hPa".format(val)
    clue_data[4].text = "Altitude: {:.1f} m".format(val)
    clue_data[5].text = "Temperature: {:.1f} C".format(val)
    clue_data[6].text = "Humidity: {:.1f} %".format(val)
    clue_data[7].text = "Proximity: {}".format(val)
    clue_data[8].text = "Gesture: {}".format("flip")
    clue_data.text_group.hidden = False
    clue_data[9].text = "Color: R: {} G: {} B: {} C: {}".format(100, 100, 100, 100)
    clue_data[10].text = "Button A: {}".format(False)
    clue_data[11].text = "Button B: {}".format(False)
    clue_data[12].text = "Touch 0: {}".format(False)
    clue_data[13].text = "Touch 1: {}".format(False)
    
    # clue_data.text_group.hidden = False
    clue_data[14].text = "Touch 2: {}".format(False)
    clue_data.show()
    print("shouldn't show until the end")
    time.sleep(0.5)
    print(clue_data.text_group.hidden)
# Create the slideshow object that plays through once alphabetically.
slideshow = SlideShow(
    board.DISPLAY,
    dwell=1,
    folder=".",
    loop=True,
    fade_effect=False,
    auto_advance=True,
    order=PlayBackOrder.ALPHABETICAL,
    direction=PlayBackDirection.FORWARD,
)
print(clue_data.text_group.hidden)
for i in range(2):
    slideshow.advance()

# clue_data.show_terminal()
```

# Checklist:


src/extension.ts
```diff
@@ -621,9 +630,9 @@ export async function activate(context: vscode.ExtensionContext) {
                                         break;
                                 }
                             } catch (err) {
-                                console.log(
-                                    `Non-JSON output from the process :  ${message}`
-                                );
+                                // if not a JSON string, it is incomplete
```
xnkevinnguyen: Is there a way to handle this without catching an error (as part of the normal flow), but instead verifying if this caracter is present "\0"

xnkevinnguyen:
Confirmed tested on MAC

src/extension.ts
andreamah: At this point in the process, the main string from the python side is already split by "\0", so it wouldn't be present. 

To be clear, only individual JSON strings are added to the incompletion buffer. this means that, if the message was 

`{ test: one } \0 { test: two`

`{test:one}` and `{test:two` would be processed separately. Only `{test:two` would be added to the incompletion buffer.

vandyliu (approved):
LGTM

src/extension.ts
xnkevinnguyen: Ok, I realize I was looking at the catch for the wrong error. It is not optimal to have the catch as part of the normal flow, but after quick look there is no easy way to detect if the string is a valid json or not

xnkevinnguyen (approved):
LGTM Awesome fix, as always. Is it possible to add a comment where we say that this part is necessary for unix systems? Someone in the future might code in windows and remove it not knowing it's necessary

src/extension.ts
andreamah: Yes, it's pretty tricky to do it without looping it having the code jumping everywhere.

src/extension.ts
```diff
@@ -621,9 +634,9 @@ export async function activate(context: vscode.ExtensionContext) {
...
```
nasadigital: Currently we assume that the error caught is the one we are looking for (Perhaps `SyntaxError` in this case?).
But that might not be always the case.
For e.g.: If a developer made a typo like:
```
messageToWebview = JSON.parse(mesage);
```
in the try block, we would put the message in the buffer even though it's a ReferenceError (`ReferenceError: mesage is not defined`).
You can check the error "type" like this to make sure we have the correct exception:
```
if (err.name === "SyntaxError") {
  ...
} else { // unhandled exception
  throw err;
}
```

nasadigital (approved):
Great catch (again)!
LGTM

src/extension.ts
andreamah: Changed!

nasadigital: This is better than what I suggested, thanks for addressing!

PR #306, opened 2020-04-08
Displayio Positioning Bug Fix

andreamah:
# Description:

### Issues:
 - Setting `group.x` or `group.y` does not have an impact on the output
 - Setting `group.x`, `group.y`, `group.scale` when the group is active does not update the screen

### Solutions:
 - Adding custom properties for `group.x`,` group.y`, and `group.scale`.
 - Setting `adafruit_display_text`'s `anchored_position `already updates `group.y` and `group.x`, so we just needed to remove consideration for anchored_position and add reference to `group.y `and `group.x` whenever we need to draw.
## Type of change

# Testing:

The example code here: https://github.com/adafruit/Adafruit_CircuitPython_CLUE/blob/master/examples/clue_spirit_level.py

(for the best result, use values between -10 and 10 for xy acceleration)

Also, the regular sensor code:
```

clue_data = clue.simple_text_display(text_scale=2)

while True:
    clue_data[0].text = "Acceleration: {:.2f} {:.2f} {:.2f} m/s^2".format(
        *clue.acceleration
    )
    clue_data[1].text = "Pressure: {:.3f} hPa".format(clue.pressure)
    clue_data[2].text = "Altitude: {:.1f} m".format(clue.altitude)
    clue_data[3].text = "Temperature: {:.1f} C".format(clue.temperature)
    clue_data[4].text = "Humidity: {:.1f} %".format(clue.humidity)
    clue_data[5].text = "Proximity: {}".format(clue.proximity)
    clue_data[6].text = "Gesture: {}".format(clue.gesture)
    clue_data[7].text = "Color: R: {} G: {} B: {} C: {}".format(*clue.color)
    clue_data[8].text = "Button A: {}".format(clue.button_a)
    clue_data[9].text = "Button B: {}".format(clue.button_b)

    clue_data.show()

```
# Checklist:


src/base_circuitpython/displayio/group.py
```diff
@@ -63,6 +63,39 @@ def __init__(
...
+    @property
+    def x(self):
+        return self.__x
+
+    @x.setter
+    def x(self, val):
+        changed = val != self.__x
+        self.__x = val
+        if changed:
+            self.__elem_changed()
```
nasadigital: _&ast;nit&ast;_
Can streamline the setters a bit by omitting `changed` like this:
```
def x(self, val):
  if self.__x != val:
    self.__x = val
    self.__elem_changed()
```
Up to you!

nasadigital (approved):
Good catch!
LGTM

src/base_circuitpython/displayio/group.py
andreamah: Ahh, true true!!

src/base_circuitpython/displayio/group.py
```diff
@@ -69,9 +69,8 @@ def x(self):
 
     @x.setter
     def x(self, val):
-        changed = val != self.__x
-        self.__x = val
-        if changed:
+        if val != self.__x:
```
nasadigital: _&ast;nit&ast;_
I would change the order here to:
`if self.__x != val:`
a tiny bit easier to read when the property is on the left and the new value is on the right.

vandyliu: agreed

vandyliu (approved):
Cool!

PR #312, opened 2020-04-09
Max input length

xnkevinnguyen:
# Description:
The input will have max number of digit according to the length of the max value.

Previous behaviour: range 1-100, max length of input was hardcoded to 4
Current behaviour: range 1-100, max length of input is 3 because length of max is 100 ( 3 digits)

## Type of change

# Limitations:

# Testing:

- [ ] Test A
- [ ] Test B

# Checklist:


vandyliu (approved):
Very nice change :)

src/view/components/toolbar/InputSlider.tsx
```diff
@@ -30,7 +30,9 @@ class InputSlider extends React.Component<ISliderProps, any, any> {
                     value={this.props.value}
                     onInput={this.handleOnChange}
                     defaultValue={this.props.minValue.toLocaleString()}
-                    pattern="^-?[0-9]{0,4}$"
+                    pattern={`^-?[0-9]{0,${
+                        this.props.maxValue.toString().length
+                    }}$`}
```
nasadigital, line 35: Instead of having the type be text and then matching it against a regex pattern, I think it makes more sense to change the type to number and use the native attributes to validate the field. Docs [here](https://www.w3.org/wiki/HTML/Elements/input/number).
Let me know if I missed something.

xnkevinnguyen: Hi David, to provide the same experience with a number field. We would have to manipulate the number. Without regex pattern nor number manipulation, the result would be similar to this source https://jsfiddle.net/danielfilho/ZNKrb/ where the typed input can be an indefinite number

nasadigital: Sorry, I don't quite understand what you mean. Some of the attributes supported are `min`, `max`, `step`. Using a combination of these you should be able to limit the range of the input.
Works for me locally.

nasadigital (approved):

Issue #163, opened 2019-11-15
Investigate issue(s) reported on the marketplace review with regards to virtual environments and pythonPath setting


iennae:
From review on the marketplace

`Extension currently does not work with python virtual environments and ignores the "python.pythonPath"  value in settings.  Therefore all python modules must be installed globally, which is something I'm not willing to do for a beta extension.`

gramster:
Related, I got this today. I wasn't using the extension but I had just restarted VS Code. I'm guessing this is because I am on MacOS and it looks at my system Python rather than my virtual environments (where I was on 3.8). This is something we need to fix but it can be postponed a little bit while the Python extension team do some reworking of the environment API which should make the fix easier. 

![Picture1](https://user-images.githubusercontent.com/2762632/75952140-beda3980-5e62-11ea-8f9e-451e568f4594.png)

iennae:
@gramster  I was able to select a python virtual environment and not install modules globally. I am on MacOS as well. It sounds like you replicated something similar to the original reporter. Do you have more information about what could be causing the problem?

andreamah:
Hi @gramster ! 

It's strange how it happened when you weren't even using the extension, but I can maybe see why you got the error although you had a virtual environment selected. Currently, the extension uses the "python.pythonPath" from the Python extension in order to determine the version to use. However, when you first open VS Code with nothing previously set in a workspace or global settings.json (usually when you don't have a .vscode workspace folder and the global settings.json doesn't have python.pythonPath pre-defined), it just uses "python", which is the system Python. At the same time, the interpreter listed in the bottom left corner upon startup is a different from the system version (likely the newest on the machine). 

![Screen Shot 2020-03-05 at 10 48 31 AM](https://user-images.githubusercontent.com/31675041/76015121-99cfe000-5ecf-11ea-92e3-ef554d040bd7.png)

Therefore, when we open the extension, it mistakenly uses the system Python from the Python extension and shows that the version is too low. We'll look into a way to see if we can work around this in a future release.

Thanks for the feedback! :)

andreamah:
Addressed in #338 . If the Python path is not specifically set, it will search for the version of Python that is under "python3" (if available). This should fix the issue 😊.

I will wait one business day and close this if there are no further comments.

andreamah:
Closing because of no further feedback.

PR #338, opened 2020-04-17
Properly resolve Python path if not set

andreamah:
# Description:
In response to #163 

Fixes to dependency process:
- When no interpreter is set, it used to ask you to choose an interpreter. Now, it will look at your system version and see if it's sufficient. If there is no system version under "python" or if that version is too low, it will try using "python3". The follow-up use of "python3" helps UNIX systems get started, since "python" usually points to Python 2.7.
- Setup now properly checks version number & dependency checker works on more versions of Python.

## Type of change

# Limitations:

People who were using a version older that Python 3.7 will need to switch. Previously, they weren't properly told to switch off.

# Testing:
Tested on VS Code Online, which runs on Linux. Also tested on local windows. In both, tested the following cases:

- no defined pythonPath in settings
- python.pythonPath defined in settings (relative path, absolute path, and invalid path)
- python version too low
- no pip


src/service/setupService.ts
```diff
@@ -159,44 +159,149 @@ export class SetupService {
...
-
+        const systemPythonVar = tryingPython3
+            ? GLOBAL_ENV_VARS.PYTHON3
+            : GLOBAL_ENV_VARS.PYTHON;
         // try to get name from interpreter
         try {
             originalpythonExecutablePath = getConfig(CONFIG.PYTHON_PATH);
         } catch (err) {
-            originalpythonExecutablePath = GLOBAL_ENV_VARS.PYTHON;
+            originalpythonExecutablePath = systemPythonVar;
```
xnkevinnguyen: camel case variable names: should be originalPythonExecutablePath

src/service/setupService.ts
```diff
@@ -159,44 +159,149 @@ export class SetupService {
...
+                            CONSTANTS.ERROR.INVALID_PYTHON_PATH,
+                            DialogResponses.INSTALL_PYTHON
+                        )
+                        .then(
+                            (installChoice: vscode.MessageItem | undefined) => {
+                                if (
+                                    installChoice ===
+                                    DialogResponses.INSTALL_PYTHON
+                                ) {
+                                    const okAction = () => {
```
vandyliu, line 253: Should we add TelemetryEventName.SETUP_DOWNLOAD_PYTHON here as well?

src/service/setupService.ts
```diff
@@ -159,44 +159,149 @@ export class SetupService {
         return pythonExecutablePath;
     };
 
-    public getCurrentPythonExecutablePath = async () => {
+    public getCurrentPythonExecutablePath = async (
+        tryingPython3: boolean = false
```
xnkevinnguyen: suggestion(nit) for var name: isTryingPython3

xnkevinnguyen (changes requested):
Amazing work! Just had minor comments

src/service/setupService.ts
```diff
@@ -159,44 +159,149 @@ export class SetupService {
...
+                } else {
+                    // otherwise, we ran the "python" system variable
+                    // and we can try python3
+                    return this.getCurrentPythonExecutablePath(true);
+                }
+            }
+        } else {
+            // should only be applicable if the user defined their own path
+
+            // fix path to be absolute
```
nasadigital, line 277: _&ast;nit&ast;_
I feel like this if block is pretty self-documenting and that the comment can be omitted.

vandyliu:
Running python 2.7 as selected interpreter, I get
Code_33cyfTPDcI
Is the red one supposed to appear?

xnkevinnguyen (approved):
Amazing as always Andrea! I have tested without env path to python on my global, it works for me and uses a python version 3.7 and up to create venv

xnkevinnguyen:
Tested on MAC OS

src/service/setupService.ts
```diff
@@ -159,40 +159,151 @@ export class SetupService {
...
+            ? GLOBAL_ENV_VARS.PYTHON3
+            : GLOBAL_ENV_VARS.PYTHON;
         // try to get name from interpreter
         try {
-            originalpythonExecutablePath = getConfig(CONFIG.PYTHON_PATH);
+            originalPythonExecutablePath = getConfig(CONFIG.PYTHON_PATH);
         } catch (err) {
-            originalpythonExecutablePath = GLOBAL_ENV_VARS.PYTHON;
+            originalPythonExecutablePath = systemPythonVar;
         }
```
nasadigital, line 177: Honestly, I would just prompt the user to pick the interpreter in the python extension if nothing is selected and call it a day.
That way you have a single source of truth, we piggyback on any improvements that the python extension might introduce in the future and we keep this method much simpler.
What do you think?

nasadigital (approved):

andreamah:
> Running python 2.7 as selected interpreter, I get
> Code_33cyfTPDcI
> Is the red one supposed to appear?

I think I fixed it. Can you pull and try again?

vandyliu:
FIXED AWESOME STUFF ANDREA :D

vandyliu (approved):
amazing as always AAA

src/service/setupService.ts
andreamah: It's just that unix systems sometimes have an issue where the interpreter lists a different version in the bottom left corner than there is actually in the settings. For example, if you don't set a python path, the value in settings defaults to "python". In the extension, it will show your newest version of Python, but "python" will point to python2.7. To mitigate the confusion of people thinking that they're on the right version while they haven't even selected an interpreter, I put this in.

andreamah: https://github.com/microsoft/vscode-python-devicesimulator/issues/163#issuecomment-595390192

xnkevinnguyen: This was requested by Stephane.

src/service/setupService.ts
xnkevinnguyen: Might be removed in future refactoring PR, will be kept for now.

src/service/setupService.ts
nasadigital: Thanks for clarifying!

Issue #344, opened 2020-04-19
CPX slide switch seems to be backwards


xnkevinnguyen:
The slide switch also seems to be backwards from this Adafruit guide.
https://learn.adafruit.com/circuitpython-made-easy-on-circuit-playground-express/slide-switch

Referencing this issue https://github.com/microsoft/vscode-python-devicesimulator/issues/342 made by @profbrady.

xnkevinnguyen:
This issue is fixed with #350 . Will be deployed to production during the week

PR #345, opened 2020-04-20
Fix state of switch for CPX

xnkevinnguyen:
# Description:
After revision of the doc, the True state of the switch should be at the left and False at the right. See https://learn.adafruit.com/circuitpython-made-easy-on-circuit-playground-express/slide-switch.

Reported here: https://github.com/microsoft/vscode-python-devicesimulator/issues/344

Repro steps:
1.Run this code
```
from adafruit_circuitplayground import cp

while True:
    if cp.switch:
        cp.pixels[0] = (0, 0, 255)
    else:
        cp.pixels[0] = (0, 0, 0)


```
2. toggle the switch
Current behaviour: The led will turn ON if the switch is on the right

Expected behaviour: The led will turn ON if the swith is on the left.

## Type of change

# Limitations:

# Testing:

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/src/adafruit_circuitplayground/express.py
+++ b/src/adafruit_circuitplayground/express.py
@@ -37,7 +37,7 @@ class Express:
             (0, 0, 0),
         ]
         self.__state[CONSTANTS.EXPRESS_STATE.RED_LED] = False
-        self.__state[CONSTANTS.EXPRESS_STATE.SWITCH] = False
+        self.__state[CONSTANTS.EXPRESS_STATE.SWITCH] = True
         self.__state[CONSTANTS.EXPRESS_STATE.TEMPERATURE] = 0
         self.__state[CONSTANTS.EXPRESS_STATE.LIGHT] = 0
         self.__state[CONSTANTS.EXPRESS_STATE.MOTION_X] = 0
--- a/src/adafruit_circuitplayground/express.py
+++ b/src/adafruit_circuitplayground/express.py
@@ -37,7 +37,7 @@ class Express:
             (0, 0, 0),
         ]
         self.__state[CONSTANTS.EXPRESS_STATE.RED_LED] = False
-        self.__state[CONSTANTS.EXPRESS_STATE.SWITCH] = True
+        self.__state[CONSTANTS.EXPRESS_STATE.SWITCH] = False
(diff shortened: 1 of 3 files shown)


src/adafruit_circuitplayground/express.py
```diff
@@ -37,7 +37,7 @@ def __init__(self):
             (0, 0, 0),
         ]
         self.__state[CONSTANTS.EXPRESS_STATE.RED_LED] = False
-        self.__state[CONSTANTS.EXPRESS_STATE.SWITCH] = False
+        self.__state[CONSTANTS.EXPRESS_STATE.SWITCH] = True
```
nasadigital: I would expect the switch to be turned off by default, would you agree?

xnkevinnguyen: You're right. After looking at the pictures, it seems that default is off.

src/view/components/cpx/CpxImage.tsx
```diff
@@ -380,7 +380,7 @@ export const updateSwitch = (switchState: boolean): void => {
     if (switchElement && switchInner) {
         svg.addClass(switchInner, "sim-slide-switch-inner");
 
-        if (switchState) {
+        if (!switchState) {
             svg.addClass(switchInner, "on");
             switchInner.setAttribute("transform", "translate(-5,0)");
         } else {
```
nasadigital, line 386: `
if (condition) {
doIfTrue();
} else {
doIfFalse();
}
` is easier to understand than `if (!condition) { doIfFalse(); } else { doIfTrue(); }`
I would swap the blocks.

nasadigital:
Just a comment, the switch is red when turned off and grey when turned on. That's fine with me, since previously it used to be red when turned on but you might want to think about it.

nasadigital (approved):

xnkevinnguyen:
> Just a comment, the switch is red when turned off and grey when turned on. That's fine with me, since previously it used to be red when turned on but you might want to think about it.

Yea, we wanted to inverse it since users were expecting the opposite. Maybe Green would be a better choice then gray for ON. But will be thought of in the future

PR #351, opened 2020-04-21
Fixed print() statements for CLUE display in debug mode

andreamah:
# Description:

Previously, print statements did not show on the CLUE's display while in debug mode. I added a wrapper for `print()` in debug_user_code.py so that prints would also be sent to the clue display.

I also added starting and ending print statements to each run to match non-debug mode.

## Type of change

# Testing:

Prints should appear both on the clue screen and in vscode's terminal.
```
import datetime
import time

clue_data = clue.simple_text_display(title="Data Screen!", title_scale=1, text_scale=3)

val = 100
for i in range(12):
    val += 1
    clue_data.show_terminal()
    time.sleep(0.5)
    print(val)
    print(f"time {datetime.datetime.now(datetime.timezone.utc).strftime('%Y-%m-%d %H:%M:%S.%f')[:-3]}")


for i in range(10):
    val += 1
    clue_data[3].text = "Count: {}".format(val)
    clue_data.show()
    print("shouldn't show until the end")
    time.sleep(0.5)
    
```
# Checklist:

### DIFF
--- a/src/debug_user_code.py
+++ b/src/debug_user_code.py
@@ -32,8 +32,14 @@ sys.path.insert(0, os.path.join(abs_path_to_parent_dir, CONSTANTS.CLUE))
 # Insert absolute path to Circuitpython libraries for CLUE into sys.path
 sys.path.insert(0, os.path.join(abs_path_to_parent_dir, CONSTANTS.CIRCUITPYTHON))
 
+# get board so we can get terminal handle
+import board
+
 # This import must happen after the sys.path is modified
 from common import debugger_communication_client
+
+# get handle to terminal for clue
+curr_terminal = board.DISPLAY.terminal
 
 ## Execute User Code ##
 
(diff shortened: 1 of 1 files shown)


sagarmanchanda (approved):
lgtm :)

src/debug_user_code.py
```diff
@@ -56,12 +62,26 @@
...
+    global curr_terminal
+
+    def wrapped_func(*args, **kwargs):
+        curr_terminal.add_str_to_terminal("".join(str(e) for e in args))
+        return func(*args, **kwargs)
+
+    return wrapped_func
+
+
+print = print_decorator(print)
```
vandyliu, line 76: do we need to restore print in the end?

andreamah: That's us using the print_decorator function to override print with our custom version of print.

vandyliu: should we restore it though after debug_user_code is ran?
so like
original_print = print
print = print_decorator(print)

...
print = original_print

andreamah: In what case would this needed? When the program is done, all globals are reset.

vandyliu: I guess so. Thanks Andrea

vandyliu (approved):
Lgtm

PR #353, opened 2020-04-21
Release Note Apr 21

andreamah:
# Description:

Looks like the following:

image

The end is the same as the previous release note.


isadorasophia (approved):

src/latest_release_note.ts
```diff
@@ -1,26 +1,39 @@
...
+</div>
+
+<h2>📝(April 21, 2020)</h2>
```
nasadigital: _&ast;nit&ast;_
Perhaps omit the brackets around the date?

nasadigital (approved):

src/latest_release_note.ts
andreamah: Updated!

Issue #367, opened 2020-05-06
Getting started tab for Circuit Playground Express references micro:bit


kattni:
The first instruction on the Getting started tab for the Circuit Playground Express is "Import the micro:bit library to use it (This is required)". 

Presumably it should say "Import the Circuit Playground library to use it (This is required)"

![GettingStartedCPX](https://user-images.githubusercontent.com/7865090/81221604-12771980-8fb1-11ea-803d-6a296d1bedcd.png)

xnkevinnguyen:
Fixed with #370

PR #369, opened 2020-05-06
Correct typo for cpx import on getting started

xnkevinnguyen:
# Description:
Minor PR to fix typo on getting started page for cpx. Remove micro:bit and replace with Circuit Playground.

Screen Shot 2020-05-06 at 6 20 41 PM

Issue referenced here #367

### DIFF
--- a/src/view/pages/__snapshots__/gettingStarted.spec.tsx.snap
+++ b/src/view/pages/__snapshots__/gettingStarted.spec.tsx.snap
@@ -59,7 +59,7 @@ exports[`GettingStartedPage component  should render correctly 1`] = `
     <h3
       className="normalFontWeight"
     >
-      &gt; Import the micro:bit library to use it (This is required)
+      Import the Circuit Playground library to use it (This is required)
     </h3>
     <span
       className="codeBox"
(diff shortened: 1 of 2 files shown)


src/view/pages/__snapshots__/gettingStarted.spec.tsx.snap
```diff
@@ -59,7 +59,7 @@ exports[`GettingStartedPage component  should render correctly 1`] = `
     <h3
       className="normalFontWeight"
     >
-      &gt; Import the micro:bit library to use it (This is required)
+      Import the Circuit Playground library to use it (This is required)
```
nasadigital, line 62: It seems that we are keeping the '>' symbol at the start of the line in other lines. Perhaps we keep it here as well? Up to you.

nasadigital (approved):

src/view/pages/__snapshots__/gettingStarted.spec.tsx.snap
xnkevinnguyen: Yep, will re-add it. im making a new PR with correct branch name

Issue #366, opened 2020-05-06
Getting started tab navigation


kattni:
This may be a limitation of VS Code.

When I open the Getting started tab, it provides a drop down menu to choose what I would like to "Learn more about:". I choose the board I am interested in, copy a code snippet, and click on my `code.py` file to paste the code snippet. When I return to the Getting Started tab it seems to reload, and return to the initial state with the drop down menu available. To get back to the code snippets, I must click on the menu and choose the board again each time.

If possible, consider allowing the Getting Started tab to remain in the chosen section so you can copy and paste multiple snippets of code without needing to choose the board from the drop down menu every time.

xnkevinnguyen:
This is a great suggestion, persisting the state is be possible. I'll assign myself to it :)

xnkevinnguyen:
should be persisting with #371

PR #371, opened 2020-05-06
Persist tutorial page when hidden 

xnkevinnguyen:
# Description:
To persist the state of the getting started webview, we can use https://code.visualstudio.com/api/extension-guides/webview#retaincontextwhenhidden from the vscode API.

 Please note that this has a performance overhead since we keep persist the state in the background. The reason why I am using this versus the alternative which is saving the state as a message and sending it back on reopen: 

1. Smoother experience, since it does not need to reload
2. Loaded Page is fairly lightweight and has no background process
3. Will save the exact position of the user on page.
4. More coherent with the description on #366

This PR addresses #366

### DIFF
--- a/src/service/webviewService.ts
+++ b/src/service/webviewService.ts
@@ -42,17 +42,17 @@ export class WebviewService {
                 </script>
                 <script ></script>
                 ${this.loadScript(
-            this.context,
-            webviewType,
-            "out/vendor.js",
-            hasDevice
-        )}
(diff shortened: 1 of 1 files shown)


nasadigital (approved):
LGTM, thanks for your contribution!

xnkevinnguyen:
> LGTM, thanks for your contribution!

Thank you for reviewing David!!!!

PR #375, opened 2020-05-26
Webview resources

xnkevinnguyen:
This PR addresses #374 . This PR includes the following changes:

-Add Content security policy for scripts with meta tag to only load vscode ressources scripts
-Load local script ressources on Webview using  Webview.asWebviewUri function instead of hardcoding vscode-resources.
-Changes have been made to the configuration of  webpack to not use eval on dev which is forbidden by the content security policy 

The changes follow this sample from the vscode extension team: https://github.com/Microsoft/vscode-extension-samples/blob/master/webview-sample/README.md

Testing scenarios :
These can be done on the html document loaded into the web view in src/service/webviewService.ts 
1. Inject inline html without the bounce hash on the html document lo should be sanitized
2. Loading a script outside of the vscode ressources shouldn’t work
The last two script blocks should be sanitized.
`         
            <body>
                <div id="root"></div>
                <script nonce="${nonce}" src=${vscodeImportPathSrc} ></script>
                <script nonce="${nonce}" src=${scriptSrc} ${attributeString} ></script>
                <script>console.log("This console.log should be sanitized")</script>
                <script src="https://www.external_link.com/external_file.js"></script> 
            </body>
`

### DIFF
--- a/src/service/webviewService.ts
+++ b/src/service/webviewService.ts
@@ -26,7 +26,7 @@ export class WebviewService {
         }
     }
 
-    public getWebviewContent(webviewType: string, hasDevice: boolean) {
+    public getWebviewContent(webviewType: string, hasDevice: boolean, panel: vscode.WebviewPanel) {
         return `<!DOCTYPE html>
               <html lang="en">
               <head>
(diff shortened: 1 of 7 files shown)


src/service/webviewService.ts
```diff
@@ -79,27 +89,23 @@ export class WebviewService {
...
+            return `${WEBVIEW_ATTRIBUTES_KEY.TYPE}=${webviewType} `;
         }
-        return `<script ${attributeString} src="${vscode.Uri.file(
-            context.asAbsolutePath(scriptPath)
-        )
-            .with({ scheme: "vscode-resource" })
-            .toString()}"></script>`;
     }
 }
+// Nonce generator taken from vscode extension samples
```
isadorasophia: Can you add the link pointing to the code for future reference? (I think you took from here: https://github.com/microsoft/vscode-extension-samples/blob/master/custom-editor-sample/src/util.ts)

isadorasophia (approved):
Thanks Kevin! :D

src/service/webviewService.ts
xnkevinnguyen: Sounds good! I'll do that before merging :)

xnkevinnguyen:
Thanks for reviewing @isadorasophia ! Seems like I can't run the pipeline nor merge anymore @andreamah

